PR #25, opened 2022-03-28
first recipe from pyodide

DerThorsten:


DerThorsten:
todos:
* [] make crosspython and compiler package compatible
* [] remove python binary
* [] remove python define hacls
* [] remove prints

PR #66, opened 2022-05-13
Add ipywidgets recipe

martinRenou:
Cutting out the ipykernel dependency for now (it's not technically needed if running in e.g. xeus-python)


martinRenou:
It looks like it does not find the widgetsnbextension one?
https://github.com/emscripten-forge/recipes/runs/6421932935?check_suite_focus=true#step:10:204

I guess it could be because widgetsnbextension is not yet published?

PR #68, opened 2022-05-13
Add xeus-python

martinRenou:


martinRenou:
If this passes, let's push a tag on xeus-python and use the official sources here before merging

PR #75, opened 2022-05-17
Install emperator from Pypi and emsdk from conda-forge

martinRenou:
This is all temporary. Once emperator is on conda-forge, it will pull emsdk automatically.


.github/workflows/build_changed_recipes.yaml
```diff
@@ -87,28 +84,23 @@ jobs:
...
         shell: bash -l {0}
         if: github.event_name == 'push'
         run: |
+          micromamba activate ci-env
           cd ${GITHUB_WORKSPACE}
-          export CONDA_EMSDK_DIR=$HOME/src/emsdk
-          pushd $CONDA_EMSDK_DIR
-          ./emsdk activate ${{matrix.emsdk_ver}}
-          source emsdk_env.sh
-          export FILE_PACKAGER=$CONDA_EMSDK_DIR/upstream/emscripten/tools/file_packager.py
```
martinRenou, line 111: We shouldn't need this FILE_PACKAGE anymore since https://github.com/emscripten-forge/emperator/pull/4

DerThorsten:
@martinRenou we should change a package to trigger at least the bld of one python pkg
(https://github.com/emscripten-forge/recipes/runs/6472738509?check_suite_focus=true#step:11:38)
it should be enough to change a single line in a recipe (ie use regex most the time, builds very quick)

martinRenou:
Ok gotcha :) Thanks

DerThorsten:
@martinRenou CONDA_EMSDK_DIR is indeed needed.
This might be compiled into the version of `cross-python` which is on quetz.

DerThorsten:
for bitfurnace, I think one could merge my changes, for boa it is a work in progress

DerThorsten:
https://github.com/emscripten-forge/recipes/pull/76 should build and upload a new version of cross python (3.10.1) which does not have these hardcoded paths in the pkg

DerThorsten:
@martinRenou seems like its working!

DerThorsten (approved):

martinRenou:
Thanks!

PR #84, opened 2022-05-19
trigger  xeus build

DerThorsten:


martinRenou:
Can you change the build number if the output changes?

DerThorsten:
actually this pr was superfluous ( I thought there was an error due to xeus, but it was some old superfluous `xeus-utils` recipe)

PR #86, opened 2022-05-19
Remove the force upload

martinRenou:
If it fails to upload, it should be because someone changed a recipe without changing the version number or build number, so it's good that it fails


DerThorsten:
jeah, agree, it should fail.
lets hope it fails with a good error msg

PR #127, opened 2022-07-25
update emscripten package

wolfv:
This also makes it easier to add new releases as we re-use the PKG_VERSION.

Will need to check what's wrong in boa -- run exports is not properly working. If we get it to work, we have a nice mutex package.


DerThorsten:
looks good to me

wolfv:
let me try to have a look and fix boa's run exports handling first :)

DerThorsten:
I merged boa's main into the boa branch we are using. 
What else do we need to land this?

wolfv:
We need to decide on a rebuild I guess, that's all

wolfv:
@DerThorsten we should also try to align with pyodide on the python library triplet thing. Would it be enough to change `wasm32-unknown-emscripten` in all places in sysconfigdata?

PR #131, opened 2022-08-01
add geos and shapely

wolfv:


DerThorsten:

DerThorsten:
closing since we have geos #1134  and #604

PR #135, opened 2022-08-02
Pyb2d

DerThorsten:


recipes/recipes_emscripten/pyb2d/recipe.yaml
```diff
@@ -0,0 +1,53 @@
...
+  build:
+    - '{{ compiler("cxx") }}'
+    - cmake
+    - ninja
+    - numpy
+  host:
+    - python
+    - pybind11
+    - numpy
+    - sqlite
```
wolfv, line 26: do you really need this?

recipes/recipes_emscripten/pyb2d/build.sh
```diff
@@ -0,0 +1,21 @@
+mkdir build
+cd build
+
+export CMAKE_PREFIX_PATH=$PREFIX
+export CMAKE_SYSTEM_PREFIX_PATH=$PREFIX
+
+
+# Configure step
+cmake ${CMAKE_ARGS} ..                                \
```
wolfv, line 9: why just cmake and not emcmake?

recipes/recipes_emscripten/pyb2d/build.sh
```diff
@@ -0,0 +1,21 @@
...
+    -GNinja                                           \
+    -DCMAKE_BUILD_TYPE=Release                        \
+    -DCMAKE_PREFIX_PATH=$PREFIX                       \
+    -DCMAKE_INSTALL_PREFIX=$PREFIX                    \
+    -DPYTHON_NUMPY_INCLUDE_DIR="$PREFIX/lib/python3.10/site-packages/numpy/core/include" \
+    -DPYTHON_MODULE_INSTALL_DIR="$PREFIX/lib/python3.10/site-packages"
```
wolfv, line 15: This should not be necessary with modern cmake stuff I believe.

recipes/recipes_emscripten/pyb2d/recipe.yaml
DerThorsten: probably not , at least not. 
   - sqlite
    - zlib
    - bzip2
    - libffi
   Copy pasted residual...

recipes/recipes_emscripten/pyb2d/build.sh
DerThorsten:
```
    cmake () {
        emcmake cmake "$@"
    }
```
in the compiler activation script

recipes/recipes_emscripten/pyb2d/build.sh
DerThorsten: ...well..the cmake I use might be not that modern / well-crafted. I had to add "DPYTHON_MODULE_INSTALL_DIR="$PREFIX/lib/python3.10/site-packages".

The problem is, that I use the python executable within CMake to figure out the python related path. But this does not work well in the cross-compilation case

wolfv: `find_package(Python3 REQUIRED COMPONENTS Development Numpy)` should set things up correctly I'd imagine.

For site-packages, conda-build/boa also sets the `$SP_DIR` variable (that would work across python versions).

DerThorsten: ok, then It might make sense that I make the changes upstream in `pyb2d` directely

PR #137, opened 2022-08-02
add pywavelets

wolfv:


wolfv:
I think the same library appears on the link libraries twice. Maybe that's all ...

PR #142, opened 2022-08-03
Speed up Matplotlib import

martinRenou:
Working on speeding up Matplotlib import. For now it has to generate the font cache at startup, I'd like to get rid of this.


martinRenou:
I'm removing the multiple Pyodide back-ends as we don't really need them (yet?)

DerThorsten:
looks good to me, and if tests pass it can't be that bad =)

PR #143, opened 2022-08-03
Add Empack file filtering config

DerThorsten:


empack_config.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+  matplotlib:
+    include_patterns:
+      - pattern: '*.so'
+      - pattern: '*.py'
+      - regex: '*dateutil-zoneinfo.tar.gz'
+   exclude_patterns:
+      - pattern: '**/tests/**/*.py'
+      - pattern: '**/tests/**/*.so'
+      - pattern: "*matplotlibrc"
```
martinRenou, line 19: We actually need this

```suggestion
```

empack_config.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - regex: '*dateutil-zoneinfo.tar.gz'
+   exclude_patterns:
+      - pattern: '**/tests/**/*.py'
+      - pattern: '**/tests/**/*.so'
+
+  matplotlib:
+    include_patterns:
+      - pattern: '*.so'
+      - pattern: '*.py'
+      - regex: '*dateutil-zoneinfo.tar.gz'
```
martinRenou, line 15: It seems this is not the latest version that fixed Matplotlib.

Can you also add `fontlist.json`? In combination with https://github.com/emscripten-forge/recipes/pull/142 we should get a decent import time improvement for matplotlib.

```suggestion
      - regex: '*dateutil-zoneinfo.tar.gz'
      - pattern: '**/matplotlib/mpl-data/**'
      - patter: '**/matplotlib/fontlist.json'
```

martinRenou (changes requested):

empack_config.yaml
DerThorsten: @martinRenou this was not yet ready for review and had a lots of errors, now it should be ready for review

empack_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - pattern: '*.py'
+      - pattern: '*dateutil-zoneinfo.tar.gz'
+    exclude_patterns:
+      - pattern: '**/tests/**/*.py'
+      - pattern: '**/tests/**/*.so'
+  matplotlib:
+    include_patterns:
+      - pattern: '*.so'
+      - pattern: '*.py'
+      - pattern: '*matplotlibrc'
```
martinRenou, line 14: Actually this one is already included implicitely in the line below (under `mpl-data`), we can remove it

```suggestion
```

empack_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - pattern: '*.so'
+      - pattern: '*.py'
+      - pattern: '*dateutil-zoneinfo.tar.gz'
+    exclude_patterns:
+      - pattern: '**/tests/**/*.py'
+      - pattern: '**/tests/**/*.so'
+  matplotlib:
+    include_patterns:
+      - pattern: '*.so'
+      - pattern: '*.py'
```
martinRenou, line 13: Should we keep these? It's already part of the "default" down the file.

empack_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - pattern: '**/tests/**/*.so'
+  scikit-learn:
+    include_patterns:
+      - pattern: '*.so'
+      - pattern: '*.py'
+      - pattern: '**/sklearn/datasets/**'
+    exclude_patterns:
+      - pattern: '**/tests/**/*.py'
+      - pattern: '**/tests/**/*.so'
+  scikit-image:
```
martinRenou, line 28: Shamelessly adding a new one :) It seems ipympl broke too recently

```suggestion
  ipympl:
    include_patterns:
      - pattern: '**/ipympl/static'
  scikit-image:
```

martinRenou (changes requested):

empack_config.yaml
DerThorsten: but we still need the *py files,right?

.github/workflows/build_recipes.yaml
```diff
@@ -99,7 +99,7 @@ jobs:
         shell: bash -l {0}
         run: |
           micromamba activate ci-env
-          python -m pip install git+https://github.com/emscripten-forge/empack@main              --no-deps --ignore-installed
```
martinRenou, line 102: How do we proceed to merge the PRs?

It seems `empack` needs `recipes` to contain the config file before we can release it, and this PR needs empack to be released.

Should we split this PR in two and include the config file only first?

empack_config.yaml
martinRenou: Yes, does it only include the matching patterns? It does not use the default ones?

.github/workflows/build_recipes.yaml
DerThorsten: @martinRenou no empack does not need this. empack just expects **some path to a config** in the cli.

so we could merge this PR with 
```
python -m pip install git+https://github.com/emscripten-forge/empack@config            --no-deps --ignore-installed
```
then merge https://github.com/emscripten-forge/empack/pull/30

then we change from `config` to `main` and make a new PR
```
python -m pip install git+https://github.com/emscripten-forge/empack@config            --no-deps --ignore-installed
```
this should then not break anything

empack_config.yaml
DerThorsten: the default is just for PKGs that are not explicitly listed. so you still need to add the stuff

DerThorsten: first, the files are checked against the include-patterns, if any of these matches, the file is considered.
All considered files are checked against all exclude patterns, if none of the exclude pattern are matching, the file will be included

martinRenou: Ok :) I updated my suggestion accordingly

DerThorsten:
I think we should also add a ci-test to check if the provided config.yaml matches the schema

.github/workflows/build_recipes.yaml
martinRenou: Would it be possible to make a non-breaking empack change?

Maybe by having a default config file pointing to the one from this PR? So https://raw.githubusercontent.com/emscripten-forge/recipes/main/empack_config.yml once this is merged

DerThorsten: the empack changes also change the CLI itself, so instead of `empack pack python core` it is `empack pack env ...` 

so I would prefer the breaking changes...

cant we make a major version increment and constraint xeus-lite to empack <1.x.x  until we also update xeus-lite?

martinRenou: Yes we should do it. I'll update xeus-python lite and make a release right away.

PR #152, opened 2022-08-05
Udunits and expat

wolfv:
@ocefpaf was interested in this :)

### DIFF
--- a/recipes/recipes_emscripten/udunits2/recipe.yaml
+++ b/recipes/recipes_emscripten/udunits2/recipe.yaml
@@ -8,6 +8,8 @@ package:
 source:
   url: https://artifacts.unidata.ucar.edu/repository/downloads-udunits/{{ version }}/udunits-{{ version }}.tar.gz
   sha256: 590baec83161a3fd62c00efa66f6113cec8a7c461e3f61a5182167e0cc5d579e
+  patches:
+    - 0001-static-path-xml.patch
 
 build:
   number: 0
--- a/recipes/recipes_emscripten/udunits2/recipe.yaml
+++ b/recipes/recipes_emscripten/udunits2/recipe.yaml
@@ -1,0 +1,41 @@
+context:
+  version: "2.2.28"
+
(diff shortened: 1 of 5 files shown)


ocefpaf:
Pinging also @dopplershift here b/c I believe he is also interested in this. Awesome work! I need to try it out with [cf-units](https://github.com/SciTools/cf-units) now.

wolfv:
can you write a small tests for that python pacakge?

wolfv:
e.g. like this: https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/gmpy2/test_gmpy2.py

dopplershift:
🎉

recipes/recipes_emscripten/udunits2/recipe.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - texinfo  # [not win]
+  host:
+    - expat
+  run:
+    - expat
+
+# test:
+#   commands:
+#     - udunits2 -h
+#     - udunits2 -H meter -W miles
```
ocefpaf, line 33: @wolfv you asked me to write a test in https://github.com/emscripten-forge/recipes/pull/152#issuecomment-1206474268 but these are commented out. What happened here? Should I try to re-enable them?

wolfv: Hey @ocefpaf sorry for the confusion. We actually use a custom test thingy because the tests need to run in a weird environment (inside NodeJS etc.).

E.g. it looks like:

```yaml
extra:
  emscripten_tests:
    python:
      pytest_files:
        - test_cftime.py
```

Where `test_cftime.py` is a file in the same folder as `recipe.yaml`. You can find the cftime example here: https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/cftime/recipe.yaml#L39-L43

If you want to add some more tests, feel free! :)

PR #154, opened 2022-08-05
Update wasm-request package

martinRenou:
- [x] TODO Add a `requests` output that pins to the polyfill


martinRenou:
@wolfv There might be versioning mismatch with this approach. If another conda package relies on `requests>=1,<2` they simply won't get it.

Do you know if we can work around this?

martinRenou:
Ideally we'd match the requests API exactly and respect the same versioning.

wolfv:
looks good to me. I don't know what to do about versioning. I guess any requests usage out of the ordinary would break pretty soon, anyways, right?

martinRenou:
Yes. For now it only provides the basic features.

martinRenou:
Thanks for the review!

PR #157, opened 2022-08-07
try llvm

wolfv:


wolfv:
We can probably disable most targets and only compile `wasm-32` and `wasm-64`.

PR #160, opened 2022-08-08
Include xarray's static files (HTML and CSS)

davidbrochart:
Fixes https://github.com/jupyterlite/jupyterlite/issues/760.


martinRenou (approved):
Thanks!

martinRenou:
jupyterlite-xeus-python downloads this file dynamically from this repo, this means that you can already retry a jupyterlite build and see if it works :)

davidbrochart:
It does work, thanks!

PR #166, opened 2022-08-08
Update setuptools

martinRenou:


empack_config.yaml
```diff
@@ -1,4 +1,10 @@
 packages:
+  setuptools:
+    include_patterns:
+      - pattern: '**/pkg_resources/*.so'
+      - pattern: '**/pkg_resources/**/*.so'
+      - pattern: '**/pkg_resources/*.py'
+      - pattern: '**/pkg_resources/**/*.py'
```
martinRenou, line 7: Not adding the `setuptools` module on purpose for now. If some people wants it we can rethink, but they can always use another config.

PR #167, opened 2022-08-08
Fix hardcoded Python version in pyerfa build script

martinRenou:


martinRenou:
Arf ok. I'll update the build number.

PR #169, opened 2022-08-08
Remove jedi recipe

### LINKED ISSUE #165: Remove jedi recipe
We should not be needing a `jedi` recipe as the conda-forge recipe should probably be noarch (see https://github.com/conda-forge/jedi-feedstock/pull/52). Let's remove the jedi recipe from here once it's noarch.

martinRenou:
The conda-forge recipe is now noarch, we don't need this anymore.

Fix #165


wolfv:
cool!

PR #193, opened 2022-09-02
patch sklearn dataset utils

DerThorsten:


martinRenou:
Maybe we should try to make `urlretrieve` work if we can?

PR #194, opened 2022-09-12
Ignore some patterns

martinRenou:
We're not actually gaining so much from ignoring those, but it can go up to one useless MB with many widgets libraries, so it's still worth doing it.


martinRenou:
I was kind-of hoping it would be more of a gain, because I wrongly thought we were including the JS for widgets extensions, which we were not doing.

PR #198, opened 2022-09-30
Add lightgbm

IsabelParedes:


wolfv:
weird, it seems that the `--nomp` is properly applied. The error log mentions that this command fails:

```
'cmake', '/home/runner/micromamba-root/envs/ci-env/conda-bld/lightgbm_1664540110182/work/compile', '-DUSE_OPENMP=OFF'`
```

recipes/recipes_emscripten/lightgbm/build.sh
```diff
@@ -0,0 +1,3 @@
+cmake ./compile -DUSE_OPENMP=OFF
+
+make -j4
```
wolfv, line 3: `-j${CPU_COUNT}` is better (e.g. in CI there is only two cores available).

PR #199, opened 2022-10-03
Add RobotRaconteur

ihuicatl:


recipes/recipes_emscripten/robotraconteur/recipe.yaml
```diff
@@ -0,0 +1,43 @@
...
+  # ldflags: -fexceptions
+  # script: build.sh
+
+requirements:
+  build:
+    - "{{ compiler('cxx') }}"
+    - cmake
+    - python
+    - numpy
+  host:
```
wolfv, line 26: you should add `python` in `host` as well

PR #200, opened 2022-10-04
Add swift-sim

ihuicatl:


recipes/recipes_emscripten/swift-sim/recipe.yaml
```diff
@@ -0,0 +1,53 @@
...
+  host:
+    - pip
+    - python
+    - numpy
+  run:
+    - python
+    - websockets
+    - numpy
+    - spatialmath-python
+    - spatialgeometry>=1.0.0
```
wolfv, line 34: I think a space between `spatialgeometry` and `>=1.0.0` is necessary

ihuicatl:
It does look like we will need `websockets` or an equivalent.

PR #203, opened 2022-10-07
Add HEIF packages

IsabelParedes:
Packages:
- libde265
- libheif
- pyheif
- ~~pillow_heif~~ (will be added in a separate PR)


wolfv:
it seems that the buidl tools finds some library from the system 

```
In file included from /usr/include/gdk-pixbuf-2.0/gdk-pixbuf/gdk-pixbuf-io.h:35:
In file included from /usr/include/glib-2.0/glib.h:30:
In file included from /usr/include/glib-2.0/glib/galloca.h:32:
```

It might be good to remove those. Maybe `rm -rf /usr/include` could help (maybe need to remove more stuff...)

ihuicatl:
> It might be good to remove those. Maybe rm -rf /usr/include could help (maybe need to remove more stuff...)

@wolfv it seems like I need more permissions to remove anything from `/usr/include` and it also looks like it's ignoring the flag `--disable-gdk-pixbuf` or that flag does not work as expected.
So instead I commented out the line where it looks for that package and that seems to do the trick.

ihuicatl:
Added `libc6-dev-i386` to fix this error:
```
fatal error: 'bits/libc-header-start.h' file not found
  #include <bits/libc-header-start.h>
  ```

PR #205, opened 2022-10-11
add cf_units

ocefpaf:
@wolfv this adds `cf_units` and should work as a "poor's man" test for udunits2 too.

### DIFF
--- a/recipes/recipes_emscripten/cf_units/build.sh
+++ b/recipes/recipes_emscripten/cf_units/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# Make sure cf_units can find the udunits library.
+SITECFG=cf_units/etc/site.cfg
+echo "[System]" > ${SITECFG}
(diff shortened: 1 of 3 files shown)


wolfv:
Cool!

ocefpaf:
@wolfv this is probably what I was afraid of, udunits2 is not functional without an activation script like https://github.com/conda-forge/udunits2-feedstock/blob/main/recipe/scripts/activate.sh. We can probably hardcode stuff. I'll investigate what the pyodide folks do in cases like these.

wolfv:
I think I also looked at this back then, and I thought it would be possible to patch the correct value into the binary.

ocefpaf:
> I think I also looked at this back then, and I thought it would be possible to patch the correct value into the binary.

Cool. I'd like to avoid what was done in pyodide b/c it leads an increase in size when we need to copy data files over to each wheel that requires them. I'm sure they are looking into better options too.

DerThorsten:

DerThorsten:
closing because of inactivity

PR #206, opened 2022-10-14
Picomamba

DerThorsten:
* added (new version of) picomamba
* cleaned up emscripten compiler package st. it does not depend on empack 
* removed packaging from CI

### DIFF
--- a/empack_config.yaml
+++ b/empack_config.yaml
@@ -80,6 +80,7 @@ packages:
       - pattern: '**/tests/**/*.so'
   scikit-image:
     include_patterns:
+      - pattern: '*.txt'
       - pattern: '*.so'
       - pattern: '*.py'
       - pattern: '*.json'
--- a/empack_config.yaml
+++ b/empack_config.yaml
@@ -69,20 +69,6 @@ packages:
     exclude_patterns:
       - pattern: '**/tests/**/*.py'
       - pattern: '**/tests/**/*.so'
-  scipy:
-    include_patterns:
-      - pattern: '*.so'
-      - pattern: '*.py'
-      - pattern: '*.json'
-    exclude_patterns:
-      - pattern: '**/tests/**/*.py'
-      - pattern: '**/tests/**/*.so'
-      - pattern: '**/tests/*.py'
-      - pattern: '**/tests/*.so'
-      - pattern: '**/stats/tests/**/*'
-      - pattern: '**/_test_*.so'
-      - pattern: '**/_ctest.*.so'
-      - pattern: '**/_cytest.*.so'
   scikit-learn:
     include_patterns:
       - pattern: '*.so'
@@ -94,7 +80,6 @@ packages:
       - pattern: '**/tests/**/*.so'
   scikit-image:
     include_patterns:
-      - pattern: '*.txt'
       - pattern: '*.so'
       - pattern: '*.py'
       - pattern: '*.json'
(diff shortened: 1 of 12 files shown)


wolfv:
Hmm, couldn't we do the check for a pre-installed emsdk in the emsdk package? And keep the dependency?

empack_config.yaml
```diff
@@ -69,6 +69,20 @@ packages:
     exclude_patterns:
       - pattern: '**/tests/**/*.py'
       - pattern: '**/tests/**/*.so'
+  scipy:
```
wolfv, line 72: shouldn't this go in a seperate PR?

recipes/recipes_emscripten/picomamba/recipe.yaml
```diff
@@ -0,0 +1,50 @@
+context:
```
wolfv, line 1: I think it would be cleaner to put the new recipe in a seperate PR

recipes/recipes/emscripten_emscripten-32/activate.sh
```diff
@@ -1,13 +1,24 @@
...
 if [ -z ${CONDA_FORGE_EMSCRIPTEN_ACTIVATED+x} ]; then
+
+
+
     export CONDA_FORGE_EMSCRIPTEN_ACTIVATED=1
 
     export EMSDK_PYTHON=${BUILD_PREFIX}/bin/python3
     export PYTHON=${BUILD_PREFIX}/bin/python3
 
     emsdk install  3.1.2
```
wolfv, line 11: How about checking for another env var here? Such as `CONDA_EXTERNAL_EMSDK` or so, and if that one is set, use that EMSDK? So that we can keep both functionalities.

wolfv: This is important if we ever want to move to conda-forge...

empack_config.yaml
DerThorsten: this is not yet a finished pr, but yes

recipes/recipes/emscripten_emscripten-32/activate.sh
DerThorsten: yeah, was thinking about the same.

But this pr is basically just trying to fix the "broken" state that each local build forces a download of the emscripten toolchain ... I consider this as a bug/broken

DerThorsten:
> Hmm, couldn't we do the check for a pre-installed emsdk in the emsdk package? And keep the dependency?

we could..but I just want to restore to a non-broken state where local builds do not require a download for each build

wolfv:
yes, exactly, we could conditionally skip the `install` or rather use the `emsdk` pointed to by the external install directory.

There are two ways:

- pass in an env var into the build. That's a bit tricky (usually needs modifiying each recipe)
- Always have teh external emsdk in a given location (e.g. `~/emsdk`). If we find it there, we skip the installation.

recipes/recipes/emscripten_emscripten-32/activate.sh
```diff
@@ -1,13 +1,24 @@
...
 
     export EMSDK_PYTHON=${BUILD_PREFIX}/bin/python3
     export PYTHON=${BUILD_PREFIX}/bin/python3
 
-    emsdk install  3.1.2
-    emsdk activate 3.1.2
-    export CONDA_EMSDK_DIR=$BUILD_PREFIX/lib/python$($PYTHON -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')")/site-packages/emsdk
+
+
+    if [ -n "$CONDA_EMSDK_DIR" ]; then
```
wolfv, line 11: cool! Does this work locally with boa? I am wondering because I thought that most environment variables are removed when boa runs.

Issue #217, opened 2022-11-14
Manual build instructions questions


rprimet:
Hi,

Since @DerThorsten provided manual build instructions (thanks!), I tried to reproduce those and ran into a few snags, and also have a few questions, so here goes:

### misc. typos (not in shell commands):
"can be build" -> can be built
aftger -> after
localy -> locally
Bootrap -> Bootstrap
popoulate -> populate

### typos in commands
(**note: I used my host install of mamba rather than micromamba, is that possible or not (or discouraged)? Some of the issues may be related to that choice**) -- maybe the answer would entail using docker rather than attempting a build directly on a developer's machine? My choice was guided by the worry that installing micromamba (which is mainly intended for CI?) alongside regular mamba might trample on the host's mamba install...

`ci_env.yaml` -> `ci_env.yml`
`conda create` -> `conda env create` 

(so the env creation command becomes `conda env create -n emforge --file ci_env.yml`)

### issues / malfunctions
- To get the bootstrap build to run, I had to explicitly set the target to emscripten-32, i.e.

`python builder.py build explicit  recipes/recipes_emscripten/bzip2    --skip-tests --emscripten-32`

- Even so, it initially failed
- After the initial failure, I added the local conda channel to my `.condarc` and ran a "full" build (without `--skip-tests`), which ran until the sanity check (run `bzip2 --help`) which failed due to `node` not being found (is there a missing step for activating the emscripten environment?)
- Running the "binary" manually using the emscripten-provided `node` works! 

```
/home/romain/.cache/empack/emsdk-3.1.2/node/14.18.2_64bit/bin/node /home/romain/mambaforge/envs/emforge/conda-bld/bzip2_1668419287752/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/bin/bzip2.js --help
bzip2, a block-sorting file compressor.  Version 1.0.8, 13-Jul-2019.

   usage: bzip2.js [flags and input files in any order]

   -h --help           print this message
   -d --decompress     force decompression
   -z --compress       force compression
   -k --keep           keep (don't delete) input files
   -f --force          overwrite existing output files
   -t --test           test compressed file integrity
   -c --stdout         output to standard out
   -q --quiet          suppress noncritical error messages
   -v --verbose        be verbose (a 2nd -v gives more)
   -L --license        display software version & license
   -V --version        display software version & license
[...]

PR #219, opened 2022-11-17
removed test from bzip2 which requires node

DerThorsten:
**outdated:**
removed test from bzip2 which requires node as testing for emscripten-32 is not reliable when non-emscripten-32 dependencies are needed (ie `node` in this case) because we rely on system libs  (https://github.com/emscripten-forge/recipes/issues/217)

### DIFF
--- a/recipes/recipes_emscripten/bzip2/recipe.yaml
+++ b/recipes/recipes_emscripten/bzip2/recipe.yaml
@@ -31,6 +31,8 @@ features:
     default: false
 
 test:
+  requires:
+    - sel(emscripten): '{{ compiler("c") }}'
   commands:
   - sel(emscripten): $CONDA_EMSDK_DIR/node/$(ls $CONDA_EMSDK_DIR/node/)/bin/node ${PREFIX}/bin/bzip2.js --help
   - sel(unix and not emscripten): bzip2 --help
--- a/recipes/recipes_emscripten/bzip2/recipe.yaml
+++ b/recipes/recipes_emscripten/bzip2/recipe.yaml
@@ -32,7 +32,6 @@ features:
 
 test:
   commands:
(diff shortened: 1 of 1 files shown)


wolfv:
It's possible to add test depenendencies (such as node)

DerThorsten:
> It's possible to add test depenendencies (such as node)

@wolfv the problem is, that the test-dependencies are for `emscripten-32` but we need node for `linux-64` or whatever machine is used to build / test the recipe

rprimet:
Since the test seems valuable, it might be better to activate the emscripten-provided `node` and use it, rather than remove the test?

DerThorsten:
> Since the test seems valuable, it might be better to activate the emscripten-provided `node` and use it, rather than remove the test?

If there is a "nice" way to do this in a conda-forge-ish fashion that would be great.
But I would prefer a more general way to set up test dependencies that are **not only** for the target platform.

This would also remove the need for the special section in the recipes for the python tests (for example here: https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/scikit-learn/recipe.yaml#L130-L133)

DerThorsten:
I changed the command to `- sel(emscripten): $CONDA_EMSDK_DIR/node/$(ls $CONDA_EMSDK_DIR/node/)/bin/node ${PREFIX}/bin/bzip2.js --help` which should use the emscripten provided `node`

DerThorsten:
@wolfv  to bad `env` variables defined in the compiler activation script are not defined in the test section any more, so
`- sel(emscripten): $CONDA_EMSDK_DIR/node/$(ls $CONDA_EMSDK_DIR/node/)/bin/node ${PREFIX}/bin/bzip2.js --help` does
not work as `CONDA_EMSDK_DIR` is not defined anymore

PR #260, opened 2023-02-01
Use tag instead of last commit for Nelson

JohanMabille:


martinRenou:
Those recipes are not being built. 

Would you like to take over https://github.com/emscripten-forge/recipes/pull/275 ?

martinRenou:
I'm fine if you just close my PR and take the changes from there

PR #279, opened 2023-03-06
debug re2

DerThorsten:


DerThorsten:
re2 failed in the cron job https://github.com/emscripten-forge/recipes/actions/runs/4343838780/jobs/7586383627#step:11:19348

It looks like it did not use emcc but gcc.
It seems that these changes fix that

DerThorsten:
also sneaking in a tiny change in the cronjob

PR #283, opened 2023-03-08
Add awkward-cpp

agoose77:
This PR adds `awkward-cpp`, a backend component to [`awkward`](https://pypi.org/project/awkward/).

I'm still figuring out emscripten-forge (the local build steps in the docs fail, probably due to a newer `boa`), so bear with me!


wolfv:
@agoose77 I pushed a small change ... let's see if it helps :)

wolfv:
hehe, yeah, there are some tiny differences! But on the upside, it's valid YAML! :)

You mean with the `EMSCRIPTEN=1`? I actually don't know of a good way, maybe @DerThorsten knows something ...

wolfv:
@agoose77 did you already compile this for pyodide? Basically, if you use cmake internally you need to call the emscripten cmake to use the proper toolchain (I think it's `emcmake cmake ... `).

Another way to fix this is to point directly to the emsccriptn CMake Toolchain file https://github.com/emscripten-core/emscripten/blob/main/cmake/Modules/Platform/Emscripten.cmake

agoose77:
Right! Yes, I already build under pyodide-build in our CI. At the moment, we need to change the rpath between platforms, and there's a workaround for emstrip that can eventually be removed.

agoose77:
@wolfv I'm trying to get my head around how this all works! Is the `build` section using conda-forge packages, whilst the `run` section uses `emscripten-forge`? Not super knowledgeable on the internals here, but curious to learn more!

wolfv:
Well, it uses conda-forge packages, yes, but in all three environments (if the conda-forge package is `noarch`).
The main difference is that in this case,

- `build` is the `linux-64` architecture. So all executable that need to "run" at build time shoudl go there
- `host` and `run` use the `emscripten-32` architecture, so everything that you need to `link` and run at runtime needs to go there. All `noarch` packages work, and `emscripten-32` packages are currently only on `emscripten-forge`. `noarch` packages can come from either the `conda-forge` project or from here.

I am not sure where `scikit-build-core` needs to go. I suspect you need it at build time, but as a noarch package it should be fine to install in the host env as well (I suspect some dependency of it might not be on emscripten-forge though).

agoose77:
Right, that makes more sense (that it's solely an arch setting)!

> I suspect some dependency of it might not be on emscripten-forge though

Ah, the solver reports:
> nothing provides make needed by scikit-build-core-0.1.3-pyh56297ac_0

So, `make` isn't available on `emscripten-forge`. I think `scikit-build-core` might actually make more sense as a `build` dependency anyway; it's not used for linking, it sits alongside `cmake` and `make`. I should update our upstream conda-forge recipe too.

wolfv:
nice, didn't even know about the `em-config` thing! Yeah, emscripten-forge is a little bit more "bare-metal" and less magic than `pyodide` :)

DerThorsten:
```
```

this is what I usualy have to add to pybind11 based builds

agoose77:
BTW is there a good way to test this locally? I followed the docs, but I *think* the boa branch maybe isn't pinned to a SHA and something broke?

Don't want to take too much of your time though; I can figure this out!

agoose77:
Hmm, the error I'm seeing locally upon `dlopen` is
```
Couldn't instantiate wasm: /tmp/tmpt1pg402o/prefix/lib/python3.10/site-packages/awkward_cpp/lib/libawkward-cpu-kernels.so 'Error: bad export type for `_ZN7awkward12UnionBuilder10fromsingleERKNS_7OptionsIJxdEEERKNSt3__210shared_ptrINS_7BuilderEEE`: undefined'
```

This library does not export this symbol; it doesn't even include the headers or link against any objects that define the symbol. Another side module does that. I'm therefore not sure why the process of loading this module raises the aforementioned exception. Does anyone have any clues? It's proving tricky to determine how emscripten builds and loads modules in comparison with the conventional approach, so I'm a bit unsure of where to look.

wolfv:
hmm yeah that sucks a bit! indeed, sometimes the loading sequence of emscripten seems to matter. Do you know in which `.so` file this is defined (demangled: `awkward::UnionBuilder::fromsingle(awkward::Options<long long, double> const&, std::__2::shared_ptr<awkward::Builder> const&)`)?

We did one "hack" in scipy where we used `cdll.LoadLibrary(...)` to make sure that some library is loaded before the rest and we didn't hit an undefined symbol. 

If we run into this more often, we should figure out if there is a more proper solution. I think it's also a bit of a lack of support for shared libs in emscripten that triggers these things.

agoose77:
Speedy reply, @wolfv! 

Yes, this comes from `libawkward.so`. Interesting, I might try that to see what happens...

wolfv:
Ah note that you might need to use teh GLOBAL stuff: https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/scipy/patches/0017-load-clapack-shared-lib.patch :)

You are still an early adopter btw. – we're ironing out some bumps!

agoose77:
Thanks @wolfv, I appreciate this is kind of bleeding edge build out at this stage.

I *think* changing the import order, even without the GLOBAL flag, *might* have improved things. I'm seeing a new error:
```
E           OSError: Could not load dynamic lib: /tmp/tmpv84jg_qo/prefix/lib/python3.10/site-packages/awkward_cpp/lib/libawkward.so
E           RangeError: WebAssembly.Compile is disallowed on the main thread, if the buffer size is larger than 4KB. Use WebAssembly.compile, or compile on a worker thread
```
I have seen the `RangeError` around before. It seems like the solution is to async load this module. I'm surprised that I don't have to do anything special for pyodide but I do here. 

I'm concerned that this is taking up a non-negligible amount of your time. Do you know of any other resources I should use to get some inputs here? I suspect the problems I'm seeing are more raw-emscripten than emscripten-forge.

wolfv:
I guess what I don't understand is why `libawkward.so` is in your `site-packages` (and not `lib`) and why it's not a static library? I think you might have fewer problems in that case. 

Usually the `empack` tool should do the precompilation step for the shared libraries so that loading them isn't an issue. Maybe @DerThorsten can help out.

So far it hasn't taken up much of my time at all, so don't worry about that :)

agoose77:
@wolfv this package builds a couple of libraries; a Python extension (pybind11), and a set of CPU kernels that perform ragged computations. The latter are loaded using ctypes from a shared object, which is why it lives in the same location as the source files. Older versions of our `awkward` package used to support linking against from C++ applications, but we retooled to move our business logic into Python. As such, there's little benefit in being able to link against our shared objects, so there's no need to put them in the conventional library search paths.

I am not aware of anything that precludes us from building a statically linked pybind11 extension module, but it would be a reasonable amount of work is that's not how we currently do things.

DerThorsten:
> Thanks @wolfv, I appreciate this is kind of bleeding edge build out at this stage.
> 
> I _think_ changing the import order, even without the GLOBAL flag, _might_ have improved things. I'm seeing a new error:
> 
> ```
> E           OSError: Could not load dynamic lib: /tmp/tmpv84jg_qo/prefix/lib/python3.10/site-packages/awkward_cpp/lib/libawkward.so
> E           RangeError: WebAssembly.Compile is disallowed on the main thread, if the buffer size is larger than 4KB. Use WebAssembly.compile, or compile on a worker thread
> ```
> 
> I have seen the `RangeError` around before. It seems like the solution is to async load this module. I'm surprised that I don't have to do anything special for pyodide but I do here.
> 
> I'm concerned that this is taking up a non-negligible amount of your time. Do you know of any other resources I should use to get some inputs here? I suspect the problems I'm seeing are more raw-emscripten than emscripten-forge.

this usually only happens if you try to dlopen a `*.so` for which the Instantiation has failed.

DerThorsten:
@wolfv I fear we also need to do the same as for clapack:
https://github.com/emscripten-forge/empack/blob/main/empack/patches.py#L5

so the shared library needs to end up in the root of the virtual filesystem (which one atm only can do with hacks on an `empack` level) and we then maybe need to dll open it

agoose77:
Thanks for the troubleshooting here everyone, it has been helpful!

If this is likely to be an ongoing problem for "weird" packages, is there scope for per-recipe metadata that can specify e.g. dll ordering as in https://github.com/emscripten-forge/empack/blob/main/empack/patches.py#L5?

PR #284, opened 2023-03-09
remove build path from python sysconfig data

DerThorsten:
Since we use a emscripten install living **outside** the conda-prefixes, the python sysconfig data contains path like:
` "CC": "/home/runner/.cache/empack/emsdk-3.1.27/upstream/emscripten/emcc"`

This has only been an esthetical problem since cross-python is configured to use `emcc` as compiler. But it seems that `pyo3`/`rust`/`setuptools_rust` based build systems (as we need for the new `cryptography` pkg) do indeed parse python sysconfig data file. As a consequence one cannot compile on any machine **not** having emscripten installed at `"/home/runner/.cache/empack/emsdk-3.1.27/upstream/emscripten/` (note the user name `runner` as this was build on githubs ci)

This PR adds a tiny script replacing things like `"/home/runner/.cache/empack/emsdk-3.1.27/upstream/emscripten/emcc"` with just  `"emcc"`

### DIFF
--- a/recipes/recipes_emscripten/python/build.sh
+++ b/recipes/recipes_emscripten/python/build.sh
@@ -60,19 +60,28 @@ if [[ $target_platform == "emscripten-32" ]]; then
 
     emmake make CROSS_COMPILE=yes -j8
 
-
-
+    # replace:
+    #  "some/long/path/containing_the_build_dir/emcc"  with "emcc"
+    #  "some/long/path/containing_the_build_dir/emar"  with "emar"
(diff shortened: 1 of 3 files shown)


martinRenou:
It may be simpler to implement this with 3 `sed`s?

DerThorsten:
> It may be simpler to implement this with 3 `sed`s?

we might need to parse the actual python code to do more complex things (as I did in a earlier version) therefore I would like to leave the python script in

martinRenou (approved):

PR #288, opened 2023-03-12
sourmash: init at 4.7.0

luizirber:
Similar PR in pyodide: https://github.com/pyodide/pyodide/pull/3635

Mostly copied deps from https://github.com/conda-forge/sourmash-minimal-feedstock instead of what I did for pyodide, do I need to bring in the "pure python" deps too (or those can be pulled from conda-forge/PyPI?)


wolfv:
Nice! I think @DerThorsten did some work recently to enable rust on emscripten forge so there is a chance :)

wolfv:
Regarding pure python deps – everything that is available as a `noarch` package on conda-forge should work fine!

DerThorsten:
@wolfv @luizirber I am working on support for rust pyo3 packages but its not yet working

luizirber:
> @wolfv @luizirber I am working on support for rust pyo3 packages but its not yet working

Nice!

Quick comment: sourmash is not a rust pyo3 package, it uses `cffi` for the bindings. It does use `maturin` as the build backend, which is more commonly used with `pyo3` bindings.

luizirber:
This has been open for some time, I updated to sourmash 4.8.4 but I didn't see any movement in the Rust support, so I don't think this will work yet =]

DerThorsten:

Also rust packages with maturin are working now

DerThorsten:
the error might be there because we use the a nightly version of the rust compiler https://github.com/tkaitchuck/aHash/issues/200

I try to have a more detailed look on monday

luizirber:
I ran `cargo update -p ahash` and got `v0.7.8` in sourmash now, as of https://github.com/sourmash-bio/sourmash/pull/3243

when we release a new version there (probably `4.8.11`) I'll update the PR here.

PR #289, opened 2023-03-13
remove bin and broken symlinks from python package

DerThorsten:
* remove broken symlinks:
    * prefix/bin contains a lot of broken links
    * prefix/lib/pkgconfig contains **only** broken links 
* also remove python binary itself as this is:
   1. completely useless
   2. completely untested
   3. the proper way is embed libpython / pyjs in the project where you need to embed python


wolfv:
hm, why does pkgconfig contain a broken link? `.pc` files could be useful for e.g. autotools based builds

DerThorsten:
@wolfv dont know why but they look very broken to me: https://beta.mamba.pm/get/emscripten-forge/emscripten-32/python-3.10.2-h_hash_24_cpython.tar.bz2. 

Lets first remove the broken things and add an issue for the missing pkg config.

DerThorsten:
I created an issue at https://github.com/emscripten-forge/recipes/issues/290

PR #297, opened 2023-03-20
no need to add conda-bld dir of environment to .condarc

DerThorsten:
### Problem

For a local build setup with testing capabilities, one had to add the `conda-bld` dir to the `.condarc`.
This is one not needed manual step

### Solution

Infer `conda-bld` dir in script and add channel directly to mamba


builder.py
```diff
@@ -29,6 +31,14 @@
 PKG_FILE_FILTER = pkg_file_filter_from_yaml(CONFIG_PATH)
 
 
+CONDA_PREFIX = os.environ.get("CONDA_PREFIX")
```
wolfv, line 34: Note that this path can be overridden by `CONDA_BLD_PATH` env var.

wolfv:
Looks good!

builder.py
DerThorsten: so you mean at some point in the build procedure `CONDA_PREFIX` becomes whatever is in `CONDA_BLD_PATH`?
This would be fine since we start this `builder` script from a conda-env (the env with the build dir of interest).
In the script, we get the variable at the top-level global scope. At this point it should not yet be changed

wolfv: Nope, just sayign you can modify (or set) the "build package folder" using `CONDA_BLD_PATH` to something.

E.g. in conda-forge, on Windows, it's set to `C:\\bld` to have a super short path.

PR #298, opened 2023-03-22
Remove ipywidgets recipe

martinRenou:
We don't need it anymore since ipywidgets does not depend on ipykernel anymore


jtpio:
Nice!

PR #300, opened 2023-03-30
Revert "Remove ipywidgets recipe (#298)"

jtpio:
Looks like we need to revert https://github.com/emscripten-forge/recipes/pull/298 if we want to continue using more recent versions of `ipywidgets` for now.

Since `ipykernel` was added back in https://github.com/jupyter-widgets/ipywidgets/pull/3749.

- [x] Re-add the `ipywidgets` recipe
- [x] Bump to `8.0.6`


martinRenou (approved):
😞

PR #316, opened 2023-04-18
Add workflow to bump versions

jtpio:
As discussed with @wolfv it would be great to have some automation on this repo to bump the versions of the packes.

This PR adds the workflow and script from https://github.com/mamba-org/boa-forge.


martinRenou:
Amazing! Thanks for opening this. 

Would be great to have an auto-merge bot as well at some point for those automatic PRs, if the CI is green

jtpio:
> Would be great to have an auto-merge bot as well at some point for those automatic PRs, if the CI is green

Sounds good. I think this could be a nice follow-up step.

determine_versions.py
```diff
@@ -0,0 +1,648 @@
...
+                raw_yaml["source"][0]["sha256"] = sha256_hash_for_version
+            else:
+                raw_yaml["source"]["sha256"] = sha256_hash_for_version
+        raw_yaml["build"]["number"] = 0
+    return raw_yaml
+
+if __name__ == "__main__":
+    import glob
+    loader = get_yaml_loader(typ="rt")
+    recipes = glob.glob("recipes_emscripten/**/recipe.yaml")
```
jtpio, line 642: Adapted the script from https://github.com/mamba-org/boa-forge/blob/d000999ea95a6b95d5f497e00479749f7a1b24c7/determine_versions.py#L642 to search in the `recipes_emscripten` folder.

determine_versions.py
```diff
@@ -0,0 +1,648 @@
+# License: BSD3
```
martinRenou, line 1: Could `bump_recipies_versions.py` be a better name for this file? As it really update the recipe.yaml files?

Also, I've seen this yesterday https://twitter.com/condaforge/status/1647933355877445633?s=20, is there a way to do something similar here?

jtpio: > Could `bump_recipies_versions.py` be a better name for this file? As it really update the recipe.yaml files?

Sure :+1: 

> Also, I've seen this yesterday https://twitter.com/condaforge/status/1647933355877445633?s=20, is there a way to do something similar here?

That would be nice. Although likely to work on conda feedstock only? Not sure which tool allows for doing something like this but if we find it we could indeed use it in this repo (maybe @wolfv knows).

martinRenou (approved):
Thanks! Let's try it

PR #318, opened 2023-04-19
use new empack stack

DerThorsten:
* use new empack stack (ie empack >= 3.0.0,  pyjs-code-runner >= 2.0.0, pyjs >-= 1.0.0) 
* This disables the node tests as we dropped node support atm (this might change, because we just need to install a fetch-like package for node to make node work again)
* there should be no changes for the users/recipe authors, besides:
   *  package might start to work since pyjs 1.0.0 uses a better way of instantiation the shared libraries  (ie in the correct order)
   * The `PREFIX` in the tests is always `/` instead of some random temporary directory

### DIFF
--- a/.github/workflows/build_all.yaml
+++ b/.github/workflows/build_all.yaml
@@ -57,6 +57,23 @@ jobs:
           playwright install
 
       ################################################################
+      # setup emsdk
+      ################################################################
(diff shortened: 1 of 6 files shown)


martinRenou (approved):
Thanks!

PR #319, opened 2023-04-19
Add missing tickbot file

jtpio:
Follow-up to #316 

Taken from https://github.com/mamba-org/boa-forge/blob/master/tickbot.yml

cc @martinRenou

### DIFF
--- a/tickbot.yml
+++ b/tickbot.yml
@@ -1,0 +1,12 @@
+name: tickbot
+channels:
+  - conda-forge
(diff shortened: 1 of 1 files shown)


martinRenou:
Trying the bot https://github.com/emscripten-forge/recipes/actions/runs/4743937025

jtpio:
It passed but did nothing.

Maybe we'll have to try locally in dry mode to double check.

PR #332, opened 2023-04-26
per recipe pr

DerThorsten:
this creates a single PR for each recipe where a new version is available

### DIFF
--- a/.github/workflows/new_versions_v2.yaml
+++ b/.github/workflows/new_versions_v2.yaml
@@ -18,3 +18,5 @@ jobs:
         shell: bash -l -eo pipefail {0}
         run: |
           python bump_recipes_versions_and_create_pr.py
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


wolfv:
You'll have the issue of having to build them in order though (with static linking etc.), right?

DerThorsten:
> You'll have the issue of having to build them in order though (with static linking etc.), right?

 I cannot follow


This is basically just to have a lot of tiny PRs instead of one PR with a lot of changes...

DerThorsten:
jeah it should replace the old, but I wanted to see if its indeed working.
So we can merge this as is, and try it out.
If its works well from the CI we can replace the other script

DerThorsten:
there is a tiny issue still...the PR title have the old instead of the new version =)

martinRenou:
Are we not missing an author for the commit? For when it's ran on the action?

DerThorsten:
> Are we not missing an author for the commit? For when it's ran on the action?

not sure, got the main idea from StackOverflow:
https://stackoverflow.com/questions/68057744/create-pull-request-with-github-action

DerThorsten:
> If bzip2 is updated in one PR, and Python is updated in another PR and they are merged Python first, then bzip2, then the Python build would still be linking to the older bzip2, right?
> 
> Might not be much of an issue, but just something to keep in mind.

I am willing to live with that risk!

PR #453, opened 2023-05-13
recipes_emscripten: add ngspice

proppy:
ported from https://github.com/conda-forge/ngspice-feedstock and https://github.com/hdl/conda-eda/tree/master/sim/ngspice

- removed unpackaged dependencies (openmp, x, readline)
- removed platform specific build
- removed outputs (not sure if they are supported by boa recipe spec)
- add emscripten prefix to build commands

/cc @stuarteberg (let me know if you don't want to be added there)

### DIFF
--- a/recipes/recipes_emscripten/ngspice/build.sh
+++ b/recipes/recipes_emscripten/ngspice/build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -x
+set -e
+
+# Used by autotools AX_PROG_CC_FOR_BUILD
+export CC_FOR_BUILD=${CC}
+
+./autogen.sh
+
+configure_args=(
+  --prefix=${PREFIX}
+  --disable-debug
+  --disable-dependency-tracking
+)
+
+emconfigure ./configure "${configure_args[@]}" --with-ngshared LDFLAGS="${LDFLAGS}"
+emmake make -j${CPU_COUNT}
+emmake make install
--- a/recipes/recipes_emscripten/ngspice/recipe.yaml
+++ b/recipes/recipes_emscripten/ngspice/recipe.yaml
@@ -1,0 +1,61 @@
+context:
+  version: "40"
+
+package:
+  name: ngspice-suite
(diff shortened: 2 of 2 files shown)


DerThorsten:
hey @stuarteberg greetings =)

stuarteberg:
@DerThorsten Small world. :-)

stuarteberg:
@proppy I'm happy to remain on the maintainers list, though I may not be of much assistance.

proppy:
curious if https://github.com/conda-forge/pyspice-feedstock/ would need to be repackaged as well, or if it can simply be taken from conda-forge (since it seems that it is pure python and calling out to `ngspice-lib` thru `cffi`).

DerThorsten:
> curious if https://github.com/conda-forge/pyspice-feedstock/ would need to be repackaged as well, or if it can simply be taken from conda-forge (since it seems that it is pure python and calling out to `ngspice-lib` thru `cffi`).

All noarch pks can be used from conda-forge

proppy:
> All noarch pks can be used from conda-forge

It doesn't look like pyspice is currently tagged as noarch: https://anaconda.org/conda-forge/pyspice/files

Does `noarch` needs to be transitive? While it doesn't build any C code, it does depends on arch specific packages (`ngspice-lib` and `cffi`)?

maybe @stuarteberg knows more?

recipes/recipes_emscripten/ngspice/build.sh
DerThorsten, line 1: You probably need to do smth like
```
# CLI
... -DCMAKE_PROJECT_INCLUDE=overwriteProp.cmake
```

```CMake
# overwriteProp.cmake
```

See https://github.com/emscripten-core/emscripten/issues/15276#issuecomment-1039349267

DerThorsten: otherwise cmake refused to build shared libraries with emscripten

DerThorsten (changes requested):

PR #561, opened 2023-06-19
Remove the pyjs recipe duplicate

martinRenou:
@DerThorsten do we still need this? The bot updated it to the latest and it's now a duplicate. Should we update the bot st. it does not touch this recipe? I don't fully understand why we needed a separate recipe in the first place


DerThorsten:
I believe we do not need the old recipe anymore

martinRenou:
The latest release of pluggy breaks our CI, I pin it for now. Are you ok with it?

DerThorsten:
> The latest release of pluggy breaks our CI, I pin it for now. Are you ok with it?

sure!

DerThorsten (approved):

PR #563, opened 2023-06-20
Add freetype

martinRenou:


martinRenou:
I think I'm not too far from having it right. 

I was thinking we could make proper recipes using the "ports" feature from embuilder. Pausing this work for now though.

DerThorsten:
emscripten-ports seems like its not supported that much anymore  https://github.com/emscripten-ports 
all of them are "Public archive"

DerThorsten:
so probably its better to use use https://github.com/freetype/freetype directly

martinRenou:
Ports live in the emscripten repositories now probably? https://github.com/emscripten-core/emscripten/blob/main/tools/ports/freetype.py

It does look maintained (most commits were from 3 months ago under the ports directory)

DerThorsten:
ok, then its fine with me

PR #566, opened 2023-06-20
Bring back testing

martinRenou:


martinRenou:
We need to wait before the CI passes on main before we can merge

PR #578, opened 2023-06-29
Tweaking bot automerge feature

martinRenou:
For now, just adding labels and reading them.

We should later use this labels reading for checking if automerge should happen.


martinRenou:
Merging st we can test it

PR #593, opened 2023-07-17
Add more tests for pandas and rollback to 1.4.3

martinRenou:


martinRenou:
I think we should merge this even though the CI is not running. We don't want to compile pandas anyway, we just want to rollback pandas and add more robust testing

PR #602, opened 2023-08-01
Remove ipywidgets recipe, take 2

martinRenou:
Since ipywidgets 7.8.0 and 8.1.0 it does not require ipykernel anymore and the dependency is replaced by the noarch comm module, we don't need a patched ipywidgets in recipes anymore


jtpio:
Looking good with a new jupyterlite build using ipywidgets 8.1.0:

image

PR #603, opened 2023-08-02
Add geos

jsignell:
I am not at all sure this is right or how to test it, but here's what I got by copying from: [conda-forge](https://github.com/conda-forge/geos-feedstock/tree/main/recipe) and [pyodide](https://github.com/pyodide/pyodide/blob/main/packages/geos/meta.yaml)

### DIFF
--- a/recipes/recipes_emscripten/geos/build.sh
+++ b/recipes/recipes_emscripten/geos/build.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+mkdir build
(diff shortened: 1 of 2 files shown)


ian-r-rose:
Thanks for beating me too this @jsignell :)

jsignell:
Yeah not at all sure it's right, but happy to push on it together!

jsignell:
Closing in favor of #604

PR #604, opened 2023-08-02
Add shapely

jsignell:
Similar to #603:  copying from: [conda-forge](https://github.com/conda-forge/shapely-feedstock/tree/main/recipe) and [pyodide](https://github.com/pyodide/pyodide/blob/main/packages/shapely/meta.yaml)


martinRenou:
I realize there was already a PR attempt adding shapely and geos in https://github.com/emscripten-forge/recipes/pull/131

Maybe you could take some inspiration from there? Probably you can ignore the changes in `.github/workflows/build_recipes.yaml` and `conda_build_config.yml` there.

Also it may be a good idea to add both recipes at the same time, that way you can test that shapely actually works with the geos you are building.

jsignell:
Oh man! I think I looked for issues but not PRs :facepalm: let me check. There was a big change in shapely 2.0 (released December 2022) that made it easier to build: https://shapely.readthedocs.io/en/stable/release/2.x.html#refactor-of-the-internals so the build process might be significantly different than it was when that PR was opened.

DerThorsten:
might work with https://github.com/emscripten-forge/recipes/pull/609

DerThorsten:
closing this since we have shapely (at least for the 3.1.73 branch)
https://prefix.dev/channels/emscripten-forge-dev/packages/shapely

PR #614, opened 2023-08-22
Always test major packages are working

martinRenou:


martinRenou:
Wanted to add sklearn but it seems the import of Scipy fails for some reason. This does not happen in a tested deployment in xeus-python. We may want to explore the reason for the fail though.

DerThorsten:
very good idea!

martinRenou:
> very good idea!

😄 I'll be less nervous with the bot merging its own PRs

PR #623, opened 2023-08-26
Add `awkward-cpp`

agoose77:


agoose77:
@DerThorsten I recall you having experience with pybind11-based packages. This PR builds locally, but the import test fails with
```
browser-main FAILED LinkError: WebAssembly.instantiate(): Import #244 module="env" function="PyLong_AsLongLong" error: imported function does not match the expected type
```

Any ideas?

DerThorsten:
the linker flag `-s WASM_BIGINT ` might be missing.

DerThorsten (approved):

PR #630, opened 2023-09-04
wip: add iminuit

agoose77:
This PR will add `iminuit`.

As of this moment, the build succeeds but a simple reproducer fails in xeus-python:
```python
import numpy as np
from iminuit import Minuit
from iminuit.cost import UnbinnedNLL
from scipy.stats import norm

x = norm.rvs(size=1000, random_state=1)

def pdf(x, mu, sigma):
    return norm.pdf(x, mu, sigma)

# Negative unbinned log-likelihood, you can write your own
cost = UnbinnedNLL(x, pdf)
m = Minuit(cost, mu=0, sigma=1)
m.limits["sigma"] = (0, np.inf)
m.migrad()  # find minimum
m.hesse()   # compute uncertainties
```


agoose77:
If anyone can suggest approaches to debug the xeus crash, I'd be much obliged!

DerThorsten:
We updated almost any package when we moved to python 3.11. 
So maybe its working now

DerThorsten:
@agoose77 seems to be working fine now!

PR #643, opened 2023-09-20
Emscripten 3.1.45 and Python 3.11 update

DerThorsten:
* use emscripten 3.1.45
* use patched emscripten version
* use python 3.11.x
* use `f-exception` vs `f-wasm-exception`
* build python with pyodide as "build_script"
* build libffi with pyodide as "build_script"
* compatible packages with pyodide (pyodide packages are technically installable (no API exposed atm))


DerThorsten:
# DO NO MERGE 
we probably want this branch to co-exist with main for a while

martinRenou:
Nice!!

PR #671, opened 2023-10-20
Attempt to add openmc

shimwell:
I'm keen to add and openmc package to emscripten-froge. I've not used emscripten-forge before but wanted to see if it would be straight forward to add openmc.


shimwell:
closing this as it looks like I have some work to do on the cmakelists file before it is ready to compile. Hope to be back

PR #727, opened 2023-12-18
Use version 16.0.6 for llvm

anutosh491:


DerThorsten:
@anutosh491 lets add that recipe under a different folder, ie smth like llvm16.  That way we can still have v17 version maintain by the bot (the bot tries to upgrade versions and build the recipes with the upgraded version)

DerThorsten (approved):

recipes/recipes_emscripten/llvm16/build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+export CMAKE_PREFIX_PATH=$PREFIX
+export CMAKE_SYSTEM_PREFIX_PATH=$PREFIX
+
+# clear LDFLAGS flags because they contain sWASM_BIGINT
+export LDFLAGS=""
+
+
+# Configure step
+emcmake cmake ${CMAKE_ARGS} -S ../llvm16 -B .       \
```
DerThorsten, line 12: this seems wrong

anutosh491: Thanks

PR #729, opened 2023-12-19
Use separate package names for llvm 16 & 17

anutosh491:


anutosh491:
Okay as the CI is being truncated because of the step being too long, I'll use 2 PRs for the same.

PR #737, opened 2024-01-09
Added xeus-cpp

anutosh491:


anutosh491:
Converting to draft atm !

recipes/recipes_emscripten/xeus-cpp/recipe.yaml
```diff
@@ -0,0 +1,49 @@
...
+  number: 0
+
+requirements:
+  build:
+    - '{{ compiler("cxx") }}'
+    - cmake
+    - make  # [unix]
+  host:
+    - nlohmann_json
+    - xeus-lite
```
anutosh491, line 22: Add upper bound
`  - xeus-lite <2.0`

recipes/recipes_emscripten/xeus-cpp/recipe.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - xeus >=3.0.5,<4.0
+    - xtl >=0.7,<0.8
+    - llvm =16.0.6
```
anutosh491, line 25: Use `CppInterOp` instead of `llvm`

mcbarton: As this is for a build of 0.3.0 then changing to CppInterOp would be unnecessary. Will be needed 0.4.0 comes out. Also CppInterOp uses llvm 17, so this change will not work if xeus-cpp 0.3.0 is not compatible with llvm 17.

vgvassilev:
I think we can close this PR now as it was superseded by #937.

Issue #782, opened 2024-01-31
Drop `requests-wasm-polyfill`?


jtpio:
Looking at the latest release of `urllib3`: https://github.com/urllib3/urllib3/releases/tag/2.2.0

It looks like we might be able to drop `requests-wasm-polyfill`, since they now seem to be supporting `emscripten` (not just Pyodide) as an environment, and `requests` depends on `urllib3`.

cc @DerThorsten who may also have some insights on this.

jtpio:
Trying this locally quickly, maybe this would first require to package `brotli`?

![image](https://github.com/emscripten-forge/recipes/assets/591645/6faf76f4-1265-4722-b724-042bc86306ae)


Using this test `environment.yml`:

```yaml
name: xeus-python-kernel
channels:
  - https://repo.mamba.pm/emscripten-forge
  - conda-forge
dependencies:
  - xeus-python
  - requests
  - urllib3 >=2.2.0
```

jtpio:
> Trying this locally quickly, maybe this would first require to package `brotli`?

So we seem to have `brotli` packaged here already: https://github.com/emscripten-forge/recipes/tree/main/recipes/recipes_emscripten/brotli

martinRenou:
The name is brotli https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/brotli/recipe.yaml#L5 not brotli-python, I guess that's the reason?

martinRenou:
It seems also some versions of urllib3 are pulling [`brotlipy`](https://github.com/conda-forge/brotlipy-feedstock), it's unclear to me what's the difference with `brotli-python`, we may want to package that too at some point

jtpio:
So `pysocks` is defined as an optional in `urllib3` here: https://github.com/urllib3/urllib3/blob/23f2287eb526d9384dddeedb6f6345e263bb9b86/pyproject.toml#L49-L51

But not in the feedstock it seems: https://github.com/conda-forge/urllib3-feedstock/blob/0bf0b37a36ce8513397f81457e316a1e1e206001/recipe/meta.yaml#L26

Maybe that's the issue?

jtpio:
> I believe we should either get rid of this trick or provide our own recipe. I'll go with with a custom recipe for now and open an issue on the upstream feedstock.

Yeah that sounds like the simplest approach for now :+1:

DerThorsten:
[...]

PR #788, opened 2024-02-01
Split brotli package into two

martinRenou:
Fix https://github.com/emscripten-forge/recipes/issues/782


jtpio:
> Fix #782

To actually fix #782, looks like we'll probably need to remove https://github.com/emscripten-forge/recipes/tree/main/recipes/recipes_emscripten/requests-wasm-polyfill as well?

martinRenou:
Let's go with this for now. At some point we may want to use a multi-outputs recipe (and make sure the update bot supports it)

jtpio:
OK thanks for looking into this!

PR #805, opened 2024-02-14
update xeus-python to use pyjs 1.6.0

DerThorsten:
update xeus-python to use pyjs 1.6.0


DerThorsten:
needs https://github.com/emscripten-forge/recipes/actions/runs/7902961525 to complete

PR #815, opened 2024-02-26
Add `segyio`

jtpio:
Related to experiments with `repo2jupyterlite`:

- https://github.com/stevejpurves/seg_tutorial/pull/1
- https://github.com/jupyterlite/repo2jupyterlite/issues/9


jtpio:
Thanks @DerThorsten !

Do you know what may be causing the CI failure?

DerThorsten:
> Thanks @DerThorsten !
> 
> Do you know what may be causing the CI failure?

So initially it says that cross python is leaking some stuff outside the cross-python env...this is why I tried without cross env.
But that did not work out.

If `segyio` is important I can put more serious effort in it and try a local build of it (debugging trough the CI is never a good idea)

jtpio:
> If `segyio` is important

The main motivation was to maybe document how folks could make their repos compatible with `repo2jupyterlite` as an alternative to Binder, in the context of https://github.com/jupyterlite/repo2jupyterlite/issues/9. Because it started from a real world use case with this repo: https://github.com/stevejpurves/seg_tutorial

I was thinking that this could serve as an example where a package can't be installed automatically with `repo2jupyterlite` / `jupyterlite-xeus`, and needs to be packages for `emscripten-forge`.

But we could use another package as an example.

DerThorsten:
I was able to build segyio in #2256   from a custom branch where I use scikit-build-core and  modified cmakelists

jtpio:
Nice, thanks!

PR #823, opened 2024-03-13
Add 0003-cmake-config-fix-library-suffix-and-prefix.patch to CppInterOp

mcbarton:
@vgvassilev Here is the patch to CppInterOp so it picks the right suffix when finding the library.

@anutosh491 @JohanMabille can you review this PR for me please?


DerThorsten (approved):

mcbarton:
@DerThorsten are you able to merge this PR please?

PR #825, opened 2024-03-13
Update xeus-python to 0.16

martinRenou:


martinRenou:
@DerThorsten are we safe with this one? Has all the pyjs -> pyjs_core imports been updated correctly?

DerThorsten:
> @DerThorsten are we safe with this one? Has all the pyjs -> pyjs_core imports been updated correctly?

the pyjs unit test pass by importing pyjs, pyjs itself then imports pyjs_core.
So I believe we should be save...(famous last words)

DerThorsten:
Actually, I am very confident as I already have a deployment with the changes at https://emscripten-forge.github.io/pyjs/try_from_py/

martinRenou:
Oups. I forgot to reset the build number. I guess this is not so harmful though?

DerThorsten:
no not a problem

PR #836, opened 2024-03-19
Prevent scheduled jobs from running on forks

IsabelParedes:
Run scheduled workflows on the main repo only to prevent cluttering in forks.


DerThorsten:
we might want to remove the scheduled workflow after all...its failing for a long time

IsabelParedes:
Remove just the build_all, right? The other 2 look green

DerThorsten:
we can just remove the workflow file `build_all.yaml`

DerThorsten (approved):

PR #843, opened 2024-03-25
Add 0005-Fix-config-for-msvc.patch & 0006-Export-symbols-in-Windows-DLL.patch patches cppinterop

mcbarton:
Adds 0005-Fix-config-for-msvc.patch & 0006-Export-symbols-in-Windows-DLL.patch patches cppinterop


mcbarton:
@anutosh491 @JohanMabille @DerThorsten can one you review this PR for me please? I am trying to update the package to have the same packages as what is in conda-forge.

PR #855, opened 2024-04-03
move to rattler-build.

DerThorsten:
* moved  "builder.py" into local package "emci". Once we completely moved to rattler build,  this tools should only be used by the ci (finding out which recipe changed and build them)
* added support for rattler-build:
   * if a recipe has a `rattler_recipe.yaml` we will build this with rattler
   * when multiple recipes have changes in a PR, it will:
      * use rattler-build if *all* changed  recipes have a `rattler_recipe.yaml`
      * use boa if *any* changed recipes **does not have** `rattler_recipe.yaml` but **all** recipes have
        a `recipe.yaml`
      * throw an error otherwise (ie for some recipes we only have `rattler_recipe.yaml` and for others we only have `recipe.yaml`
  
* open questions:
   * is the `[conda_build_config.yaml](https://github.com/emscripten-forge/recipes/blob/main/conda_build_config.yaml) used? 
   * if [conda_build_config.yaml](https://github.com/emscripten-forge/recipes/blob/main/conda_build_config.yaml) is used, why do we need the `variant_config.yaml` to specify emscripten as a compiler 

* todos:
   * fix recipe update bot, st. `rattler_recipe.yaml` are considered. [done]
   * bump build number of all touched recipes.                                   [not needed]


wolfv:
Congrats on getting this in!

PR #872, opened 2024-04-11
Update dependencies for LFortran

anutosh491:
`bison `: required by `build0.sh`. We have a tarball so this should not be needed
`re2c`: required by `build0.sh`. We have a tarball so this should not be needed
`python`: again should not be needed. Used by `build0.sh`
`toml`: we do not run tests for the wasm build, so should not be needed.

Also I think when we have the `LFortran_Build_to_wasm `as `ON`, we naturally have the `WITH_XEUS `option as `OFF`, so in that case we would not need any of `nlohmann_json`, `xeus-lite`, `xeus`, `cppzmq`

### DIFF
--- a/recipes/recipes_emscripten/lfortran/recipe.yaml
+++ b/recipes/recipes_emscripten/lfortran/recipe.yaml
@@ -19,11 +19,18 @@ requirements:
     - "{{ compiler('cxx') }}"
     - cmake
     - make   # [unix]
-    - zlib =1.2.13
-    - lfortran =0.34.0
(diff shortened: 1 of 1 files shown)


certik (approved):
Something like this should work.

Shaikh-Ubaid:
I am happy to be listed as a maintainer for this recipe. Feel free to add me in.

certik:
Thanks @Shaikh-Ubaid, that would be a great help. You can submit a separate PR to add you in.

PR #882, opened 2024-04-12
Add ncurses

IsabelParedes:
A dependency for R.

### DIFF
--- a/recipes/recipes_emscripten/ncurses/build.sh
+++ b/recipes/recipes_emscripten/ncurses/build.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# Get an updated config.sub and config.guess
+# Running autoreconf messes up the build so just copy these two files
+cp $BUILD_PREFIX/share/libtool/build-aux/config.* .
+
+# Required to avoid cross-compiling 'make_hash' and 'make_keys'
+# which are temporary applications used in the build process
+export BUILD_CC=$(which gcc)
+export BUILD_CFLAGS="-Wno-sWASM_BIGINT"
+export BUILD_LDFLAGS="-Wno-sWASM_BIGINT"
+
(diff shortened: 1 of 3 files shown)


DerThorsten:
@IsabelParedes there is some dedicated section on crosscompiling mentioning the `make_hash` function

https://github.com/gittup/ncurses/blob/gittup/INSTALL#L1710

IsabelParedes:
Thanks @DerThorsten ! I had missed this. Following the INSTALL and config instructions, I set `BUILD_CC` to `gcc` to compile `make_hash` and `make_keys`. It seems that that worked. Now I just need to fix the install.

IsabelParedes:
These are all JS files:

 - bin/clear
 - bin/infocmp
 - bin/ncurses6-config
 - bin/reset
 - bin/tabs
 - bin/tput
 - bin/tset

Not sure if the file extension will be an issue. Leaving a comment in case it comes up later.

DerThorsten:
> These are all JS files:
> 
> * bin/clear
> * bin/infocmp
> * bin/ncurses6-config
> * bin/reset
> * bin/tabs
> * bin/tput
> * bin/tset
> 
> Not sure if the file extension will be an issue. Leaving a comment in case it comes up later.

I prefer to either have the **bin** files removed or we test them. 
Because often they do not work as often the *.wasm file is not installed.
You could try to add a test in the test section where you try to run the binaries with node

IsabelParedes:
Hi @DerThorsten, I added the *.wasm files to bin/ and I added a test with nodejs for one of the binaries. This seems to work ok. Should I add more tests?

DerThorsten:
> Hi @DerThorsten, I added the *.wasm files to bin/ and I added a test with nodejs for one of the binaries. This seems to work ok. Should I add more tests?

lets just merge it

PR #911, opened 2024-04-29
setup with pixi

DerThorsten:
set everything up with pixi.

```bash
# installs emsdk, builds compiler pkg localy
pixi run setup                                          
  
# build a pkg for emscripten-wasm32                                                            
pixi run build-emscripten-wasm32-pkg recipes/recipes_emscripten/regex 
```


wolfv:
Looking good!

jtpio:
Should the README be updated to recommend using Pixi for the local setup?

DerThorsten:
@jtpio jeah good suggestion!

PR #918, opened 2024-05-07
convert astropy and add a test

wolfv:


DerThorsten:
thanks for reporting this

DerThorsten:
using #552

PR #926, opened 2024-05-08
remove boa support!

DerThorsten:
* remove capabilities to build with boa, only rattler-build is supported
* renamed all boa recipes from `recipe.yaml` to  `recipe_legacy.yaml`
* renamed all rattler recipes from `recipe_rattler.yaml` to  `recipe_legacy.yaml` 
* once this PR is merged, all new and existing PRs need to update to / use the rattler recipe format
* there are many recipes with an untested  rattler recipe yet. This is fine for the moment,  we can fix the broken once, when a version is updated from the bot


recipes/recipes_emscripten/lpython/recipe.yaml
```diff
@@ -1,48 +0,0 @@
...
-  license_family: BSD
-  license_file: LICENSE
-  summary: Modern interactive LLVM-based Python compiler
-  dev_url: https://github.com/lcompilers/lpython.git
-
-extra:
-  recipe-maintainers:
-    - certik
-    - anutosh491
-    - Shaikh-Ubaid
```
anutosh491, line 48: Wanted to look at lpython's recipe for something but realized the file was deleted :(
Is there a reason this was done @DerThorsten ?

If not I could raise a PR fixing this cause that only leaves us with the `build.sh` file and the `patches` dir in the lpython folder.

DerThorsten: No, that was not on purpose, feel free to re-add it.

Sorry :/

anutosh491: No issues :)

PR #1012, opened 2024-05-15
Lfortran

DerThorsten:


recipes/recipes/lfortran/recipe.yaml
```diff
@@ -0,0 +1,38 @@
...
+build:
+  number: 1
+
+requirements:
+  build:
+  - ${{ compiler('cxx') }}
+  - python
+  - re2c
+  - bison ==3.4
+  - llvmdev ==11.0.1
```
IsabelParedes, line 24: It builds with llvmdev v15 also. But after that, newer versions moved header files around and it breaks.

DerThorsten:
@IsabelParedes could you figure out on what exact commit you are? I would like to try this out with the exact version you used

IsabelParedes:
@DerThorsten lfortran v0.36.0 is out now, this new version should support `-DWITH_TARGET_WASM`

PR #1047, opened 2024-05-20
Add Include/Exclude Patterns for `pint` Package to `empack_config`

michaelweinold:
This follows advice by @martinRenou provided in 

https://github.com/jupyterlite/xeus-python-kernel/issues/79#issuecomment-1244990930

and should close 

https://github.com/jupyter-xeus/xeus-python/issues/603

(and by extension)

https://github.com/brightway-lca/brightway-live/issues/52


martinRenou (approved):
Thanks!

DerThorsten:
@martinRenou I think this is only here for historical reasons. It should be removed I belive

PR #1088, opened 2024-06-03
Updated lfortran to 0.36.0

anutosh491:


IsabelParedes:
Hi @anutosh491 ! It looks like the *lfortran.wasm* file is not packaged. 
```
Files in package:
 │ │   - bin/lfortran.js
 │ │   - include/lfortran/impure/lfortran_intrinsics.h
 │ │   - lib/liblfortran_runtime.a
 │ │   - lib/liblfortran_runtime_static.a
 ```
 
 Maybe we can try copying it manually to `$PREFIX/bin` in the build script. And then change the test to:
 
 ```yaml
 tests:
- script:
    - test -f ${PREFIX}/liblfortran_runtime.a
    - node ${PREFIX}/bin/lfortran.js --version
  requirements:
    build:
      - nodejs
```

recipes/recipes_emscripten/lfortran/recipe.yaml
```diff
@@ -33,6 +33,13 @@ requirements:
   host:
   - zlib
 
+tests:
+  - script:
+    - node ${PREFIX}/bin/test_lfortran.js
+    requirements:
+      build:
+        - nodejs
```
DerThorsten, line 41: I think nodejs should be a **run** requirement not a build requirement of the tests

DerThorsten: aahh no..wait..it must be a build requirement since we are cross compiling...sorry....

sorry for the confusion and wasted time

IsabelParedes:
Btw, this version works with `-DWITH_TARGET_WASM=yes`. It just needs
```
export EMSDK_PATH=${EMSCRIPTEN_FORGE_EMSDK_DIR}
```

anutosh491:
Cool, let's try that out !

DerThorsten:
> Btw, this version works with `-DWITH_TARGET_WASM=yes`. It just needs
> 
> ```
> export EMSDK_PATH=${EMSCRIPTEN_FORGE_EMSDK_DIR}
> ```

but this is the "to wasm compiled version with lfortran".
This is not the compiler which we use to compile fotran to wasm (this would be the recipe I tried to add here https://github.com/emscripten-forge/recipes/pull/1012)

So I think we do not need the `-DWITH_TARGET_WASM=yes`

recipes/recipes_emscripten/lfortran/recipe.yaml
```diff
@@ -33,6 +33,14 @@ requirements:
   host:
   - zlib
 
+tests:
+- script:
+    - test -f ${PREFIX}/liblfortran_runtime.a
```
IsabelParedes, line 38: This should be under `lib/`

```
- test -f ${PREFIX}/lib/liblfortran_runtime.a
```

anutosh491:
cc @IsabelParedes 
I think this is ready. We should use the `test_lfortran.js` file cause that's exactly what LFortran does in their [CI](https://github.com/lfortran/lfortran/blob/ff4486db3fd595c2dee008ec8558b7054a16ec1f/.github/workflows/CI.yml#L282C1-L289C80)
```
      - name: Test built lfortran.wasm
        shell: bash -l {0}
        run: |
            set -ex
            source $HOME/ext/emsdk/emsdk_env.sh # Activate Emscripten
            which node
            node -v
            node --experimental-wasm-bigint src/lfortran/tests/test_lfortran.js
```
The `--experimental-wasm-bigint` is only required if the node version is below `16` (Lfortran uses 14.xx)

This is the result on their master branch and we get the same 
```
===============================================================================
[doctest] test cases:  49 |  49 passed | 0 failed | 0 skipped
[doctest] assertions: 456 | 456 passed | 0 failed |
[doctest] Status: SUCCESS!
```

IsabelParedes (approved):
Looking good!

DerThorsten:
Why is `source $HOME/ext/emsdk/emsdk_env.sh # Activate Emscripten` needed.
Since this is **already compiled to wasm** there should be no need to use the
emsdk for anything

anutosh491:
> Why is source $HOME/ext/emsdk/emsdk_env.sh # Activate Emscripten needed.

Ahh that's just reference code from LFortran's CI . Haven't used it for the recipes :)

DerThorsten (approved):

PR #1091, opened 2024-06-04
Update numpy to 1.26.2

ianthomas23:
The latest numpy we have is 1.26.0. The build for numpy 1.26.1 (#2786) doesn't work for me locally or in CI, but the trivial bumping of the tarball version in this PR works for me locally and the same goes for later 1.26.x releases, so perhaps if this works in CI we can just skip 1.26.1.


DerThorsten:
nice, jeah lets just skip it.

Thanks!

PR #1120, opened 2024-06-14
Add Gtest

KGB99:
is the gmock package needed here?

### DIFF
--- a/recipes/recipes_emscripten/gtest/build.sh
+++ b/recipes/recipes_emscripten/gtest/build.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -ex
+
+# Build and install dynamic library
+mkdir build
+cd build
+
+cmake ${CMAKE_ARGS} \
+  -GNinja \
+  -DCMAKE_INSTALL_PREFIX=$PREFIX \
+  -DCMAKE_CXX_STANDARD=17 \
+  -DBUILD_SHARED_LIBS=ON \
+  -DCMAKE_INSTALL_LIBDIR=lib \
+  -DCMAKE_BUILD_TYPE=Release \
+  ..
+ninja install
--- a/recipes/recipes_emscripten/gtest/build.sh
+++ b/recipes/recipes_emscripten/gtest/build.sh
@@ -2,7 +2,9 @@
 set -ex
 
 # Build and install dynamic library
-mkdir build
+if [ ! -d "build" ]; then
+  mkdir build
+fi
 cd build
 
 cmake ${CMAKE_ARGS} \
--- a/recipes/recipes_emscripten/gtest/build.sh
+++ b/recipes/recipes_emscripten/gtest/build.sh
@@ -2,16 +2,14 @@
 set -ex
 
 # Build and install dynamic library
-if [ ! -d "build" ]; then
-  mkdir build
-fi
+mkdir -p build
 cd build
 
-cmake ${CMAKE_ARGS} \
+emcmake cmake ${CMAKE_ARGS} \
   -GNinja \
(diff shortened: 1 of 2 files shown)


recipes/recipes_emscripten/gtest/recipe.yaml
```diff
@@ -0,0 +1,60 @@
+context:
+  version: 1.14.0
+  name: gtest-split
```
IsabelParedes, line 3: Since it's only one package, you can rename this to simply `gtest`

recipes/recipes_emscripten/gtest/recipe.yaml
```diff
@@ -0,0 +1,60 @@
...
+    run_exports:
+      - ${{ pin_subpackage('gtest', max_pin='x.x.x') }}
+  tests:
+    - script:
+      - test -f $PREFIX/lib/libgtest.a
+      - test -f $PREFIX/lib/libgmock.a
+      - test -f $PREFIX/include/gtest/gtest.h
+      - test -f $PREFIX/include/gmock/gmock.h 
+
+# empty wrapper for compatibility with previous output naming/separation
```
IsabelParedes, line 37: You can remove this `gmock` section all together.

recipes/recipes_emscripten/gtest/recipe.yaml
```diff
@@ -0,0 +1,60 @@
...
+  version: ${{ version }}
+
+source:
+  url: https://github.com/google/googletest/archive/refs/tags/v${{ version }}.tar.gz
+  sha256: 8ad598c73ad796e0d8280b082cebd82a630d73e73cd3c70057938a6501bba5d7
+
+build:
+  number: 0
+
+outputs:
```
IsabelParedes, line 16: No need for outputs either (single package) or the build script section.

recipes/recipes_emscripten/gtest/build.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ ! -d "build" ]; then
+  mkdir build
+fi
+cd build
+
+cmake ${CMAKE_ARGS} \
+  -GNinja \
+  -DCMAKE_INSTALL_PREFIX=$PREFIX \
+  -DCMAKE_CXX_STANDARD=17 \
+  -DBUILD_SHARED_LIBS=ON \
```
IsabelParedes, line 14: When you go through the build logs, you'll see warnings like this:
```
ADD_LIBRARY called with SHARED option but the target platform does not
support dynamic linking.  Building a STATIC library instead.  This may lead
to problems.
```
Prefer using static libraries. So change this to:
```
-DBUILD_SHARED_LIBS=OFF \
```

recipes/recipes_emscripten/gtest/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -ex
+
+# Build and install dynamic library
+if [ ! -d "build" ]; then
```
IsabelParedes, line 5: You don't need to change this, but just for future reference, you can also do this in a single line
```
mkdir -p build
```
which will show no error if the directory already exists.

recipes/recipes_emscripten/gtest/build.sh
```diff
@@ -0,0 +1,18 @@
...
+  mkdir build
+fi
+cd build
+
+cmake ${CMAKE_ARGS} \
```
IsabelParedes, line 10: Change to `emcmake cmake`

IsabelParedes (changes requested):

IsabelParedes (approved):

PR #1133, opened 2024-06-21
fix comparison in `build` section

wolfv:
cc @DerThorsten @h-veterinari


recipes/recipes/pytester/recipe.yaml
DerThorsten, line 1: I think for this package  the changes are wrong. We need playwright at runtime.

wolfv:
thanks @DerThorsten ! :)

PR #1134, opened 2024-06-24
Change Geos to be shared

KGB99:


DerThorsten:
This installs the shared libraries and probably some links with different names
```
 │ │ -- Installing: $PREFIX/lib/libgeos.so.3.12.2
 │ │ -- Installing: $PREFIX/lib/libgeos.so
 │ │ -- Installing: $PREFIX/lib/libgeos_c.so.1.18.2
 │ │ -- Installing: $PREFIX/lib/libgeos_c.so.1
 │ │ -- Installing: $PREFIX/lib/libgeos_c.so
```
I think we should only install other actual shared libraries because the symlinks will make some problems at a later stage...shared libraries are somewhat complicated to handle at runtime.

PR #1147, opened 2024-06-26
Add recipe for lpython

anutosh491:


anutosh491:
Actually I don't think we would need the patch anymore. Maybe I can make an lpython release, remove the patch and try again !

anutosh491:
cc @DerThorsten 
The recipe is ready.

PR #1148, opened 2024-06-26
rewrite quetz upload action

atrawog:
This PR includes a rewrite of the quetz package upload action to beta.mamba.pm.

It's using `curl` instead of `quetz-client` and adds some basic HTTP status codes checks to report a successful upload and provide some basic info about what's actually gone wrong in cases like duplicate package uploads or missing permissions on beta.mamba.pm.


DerThorsten:
the error is expected...to fix this I need to upload a new version of a package which pins pytest<8

DerThorsten (approved):

PR #1156, opened 2024-06-27
Fix pydantic-core package name

martinRenou:
To match the name of the package on conda-forge https://github.com/conda-forge/pydantic-core-feedstock/blob/main/recipe/meta.yaml#L4


martinRenou:
@DerThorsten it seems the test file is not found anymore after the rename:

```
 │ │ Finding all test files in the current directory
 │ │ Test files found:
 │ │ []
```

Do you have a clue what could be the issue?

DerThorsten:
I would bet this is still using pytest >=8

let me investigate

DerThorsten:
I am investigating...

martinRenou:
the weird thing is that [the pydantic-core rebuild I triggered this morning](https://github.com/emscripten-forge/recipes/actions/runs/9658585436/job/26743978530) passed just fine

DerThorsten:
there was a rattler build release 20h ago...not sure if this means something ;)

martinRenou:
Thanks for investigating @DerThorsten !

I'm happy with merging this if you don't mind having the pin on rattlerbuild for now

DerThorsten:
I want to test with one more rattler-build version, then we can just merge it

DerThorsten:
@martinRenou so on quetz this will appear as a "new" package  with the new name, unrelated to the old one right?

martinRenou:
Yes that's fine, we could remove any trace of `pydantic_core` there, but if someone relies on it let's keep it. I guess that would be harmless.

PR #1111, opened 2024-06-12
Add Cartopy

KGB99:

PR #1160, opened 2024-07-01
Add Proj

KGB99:


KGB99:
Needed to build cartopy #1111

DerThorsten:
@KGB99 I see you explicitly enable the build of the shared library, yet we only build a static one.

https://github.com/emscripten-forge/recipes/actions/runs/9747241725/job/26899471477#step:9:204
```
 CMake Warning (dev) at src/lib_proj.cmake:377 (add_library):
 │ │   ADD_LIBRARY called with SHARED option but the target platform does not
 │ │   support dynamic linking.  Building a STATIC library instead.  This may lead
 │ │   to problems.
 ```
 You could enable the shared libraries via some flags (
 See https://emscripten-forge.org/development/adding_packages/#cmake for details)
 
 But if you do not need a shared library, a static one would be prefered

PR #1161, opened 2024-07-01
add Pyproj

KGB99:


KGB99:
Needed to build #1111

DerThorsten:
Not sure if this was a good idea but since the build script asked for the proj binary I added proj to the build dependencies.
This then failed because it tried to link against the shared library from the build dependencies instead of the static library from the host dependencies.

So now it fails because the -R flag is somehow passed to the linker, and wasm-ld does not support this options (R is some runtime library path).

DerThorsten:
closed via #1170

PR #1191, opened 2024-07-12
fix missing .wasm for lua binaries

DerThorsten:
Added missing *.wasm file for the lua binaries (lua,luac).
It might be nice to try the Lua binary in cockle @ianthomas23


ianthomas23:
@DerThorsten Will do!

Issue #1193, opened 2024-07-12
Package `orjson`


jtpio:
It would be nice to have `orjson` available on emscripten forge.

For reference it seems to be available in Pyodide: https://github.com/pyodide/pyodide/tree/main/packages/orjson

PR #1194, opened 2024-07-12
Add `orjson`

jtpio:
Fixes https://github.com/emscripten-forge/recipes/issues/1193

Recipe based on the `pycrdt` one.


jtpio:
Ah it looks like CI is passing, so marking as ready!

DerThorsten (approved):

PR #1205, opened 2024-07-14
add qhull and update matplotlib

wolfv:
Adding qhull seemed pretty easy. 
Matplotlib - unfortunately - does not seem to work yet in Jupyterlite.

The error I am getting is:

```
LinkError: WebAssembly.instantiate(): Import #190 "env" "time": imported function does not match the expected type
```


DerThorsten:
often a missing `-sWASM_BIGINT` give erros like "imported function does not match the expected type" @wolfv

ianthomas23:
From version 3.9.0 Matplotlib switched to using `meson`, and I think more changes are required here for this. For example, there is no `setupext.py` any more, and it probably needs a meson cross-compilation file (as other `meson`-using packages do here). I have a WIP branch https://github.com/emscripten-forge/recipes/tree/matplotlib_3.9.0 that builds but fails the tests (probably because the build is wrong), and you are welcome to take any of that to use here.

wolfv:
Thanks @DerThorsten - with WASM_BIGINT it seems to work like a charm (for the two things I tested at least). Screenshot attached.

Screenshot 2024-07-15 at 18 49 36

wolfv:
3D triangulation stuff also seems to work?! 

Screenshot 2024-07-15 at 18 53 23

DerThorsten:
> 3D triangulation stuff also seems to work?!
> 
> Screenshot 2024-07-15 at 18 53 23

nice! 
Is this ready for review/merging?

wolfv:
I think so! :)

wolfv:
Although I think it can be simplified some more ...

wolfv:
Ok, so I spent a little time on simplifications but it didn't completely pan out. 

I tried to remove `numpy` and `pybind11` from the `build` dependencies. 

Instead we configure the location in the meson cross file. 

But _note_: the cross-python package does some magic upon activation (it moves all headers and shared objects in the `$BUILD_PREFIX`). So the "final" location of the numpy include dir is _still_ in the $BUILD_PREFIX).

recipes/recipes_emscripten/matplotlib/emscripten.meson.cross
```diff
@@ -0,0 +1,18 @@
+# binaries section is at the end to be able to append the python binary.
```
ianthomas23, line 1: This and the next line could be removed as it is no longer true.

wolfv: removed!

DerThorsten (approved):

ianthomas23 (approved):

PR #1219, opened 2024-07-16
Add libevalhyd

fbourgin:
Attempting to add libevalhyd based on xtensor


DerThorsten:
The CI might be broken due to some major version bump for mamba, I will fix this soon

PR #1223, opened 2024-07-17
ssl draft

DerThorsten:


DerThorsten:
`ImportError: dynamic module does not define module export function (PyInit__ssl)`

not sure what might be missing. Maybe @wolfv  has an idea.

The package is an un-vendored python library normally included directly in the python distribution.
I tried to copy what pyodide is doing https://github.com/pyodide/pyodide/blob/main/packages/ssl/meta.yaml

PR #1225, opened 2024-07-17
refactor python 3.11.3

KGB99:


KGB99:
the make step passes but the test does not, something might not be linking correctly

KGB99:
note the libffi library is being downloaded currently in the makefile, unless it is necessary for the build it could make sense to remove it

KGB99:
i didn't get the cpp test to work so i added some existence tests instead

DerThorsten:
@KGB99 looks good to be. Maybe at another "existence" test for the python library, then we can merge

DerThorsten (approved):

recipes/recipes_emscripten/python/recipe.yaml
```diff
@@ -1,40 +1,30 @@
...
 source:
-- url: https://github.com/pyodide/pyodide/archive/refs/tags/${{pyodide_version}}.tar.gz
-  sha256: c8350c4780be44fe763a66f5024197025b0f8740cc4a7c9ad833a89c69916276
+- url: https://www.python.org/ftp/python/${{ version }}/Python-${{ version }}.tar.xz
+  sha256: 8a5db99c961a7ecf27c75956189c9602c968751f11dbeae2b900dbff1c085b5e
 
 build:
-  number: 24
-  string: h_hash_24_cpython
+  number: 0
```
DerThorsten, line 13: we need to bump the existing bld nummer since its the same version (ie has to be 25)

KGB99:
sorry i accidentally force pushed a rebase and removed your commit, i added it back now

PR #1231, opened 2024-07-22
modularized binaries

DerThorsten:


DerThorsten:
I think we must use `MODULARIZE` https://emscripten.org/docs/getting_started/FAQ.html#can-i-use-multiple-emscripten-compiled-programs-on-one-web-page

PR #1234, opened 2024-07-23
refactor libffi and upgrade to 3.4.6

KGB99:


KGB99:
since libffi 3.4.5 there is support for wasm32 which helped with this refactor easier:)

DerThorsten (approved):

PR #1240, opened 2024-07-29
emscripten 3.1.64

DerThorsten:
Emscripten 3.1.64 

Note that we will still use 3.1.45 since its pinned here:

https://github.com/emscripten-forge/recipes/blob/main/conda_build_config.yaml#L28
https://github.com/emscripten-forge/recipes/blob/main/conda_build_config.yaml#L40


DerThorsten:
https://github.com/emscripten-forge/recipes/pull/1241

PR #1241, opened 2024-07-29
Update emscripten

DerThorsten:


DerThorsten:
note that this does not change the compiler version we use by default.
This just makes the 3.1.58 version available.

We use this version in a different branch

PR #1253, opened 2024-07-31
update python to 3.12.4

KGB99:


KGB99:
while python 3.11 linked to $BUILD_PREFIX/bin/python3.11 in the previous refactor, there is no proper link to python3.12 with the command `which python3.12`...

DerThorsten:
@KGB99 awesome.
Next we need to build pybind11 and pyjs with the new python build. 
The pybind11 recipe might have some hard coded python version somewhere...pyjs probably also...

PR #1258, opened 2024-08-03
Add `symengine` and `python-symengine`

richardotis:
This PR adds recipes for the C++ symbolic library [SymEngine](https://github.com/symengine/symengine) and its [Python wrapper library](https://github.com/symengine/symengine.py). The package names here follow the convention from conda-forge.

cc @isuruf


DerThorsten:
thanks for adding the recipe!
A test for the python recipe would be nice (can be enough to just import the build package)

PR #1266, opened 2024-08-07
Rebuild pytest without tests and update python in cross-python

KGB99:


KGB99:
i think we need to keep the pyjs rebuild seperate from pytest due to the build order

PR #1276, opened 2024-08-13
rebuild numpy with python 3.12 and emscripten 3.1.58 and pin pytest to >= 8

KGB99:


KGB99:
tests should run when #1268 is merged

DerThorsten:
@KGB99 I think we need to bump the build number of pytester

PR #1278, opened 2024-08-14
Use separate WASM and JS files for cockle

ianthomas23:
[Cockle](https://github.com/jupyterlite/cockle) (JupyterLite terminal) now supports dynamic loading of Emscripten-built WASM modules provided the WASM is embedded in a JavaScript file. With just a few adjustments this should be able to support separate WASM and JS files. To complete that I need to build the 3 modules (`cockle_fs`, `coreutils` and `grep`) with the ~`-sMODULARIZE=1`~ `-sSINGLE_FILE=1` flag removed, which is what this PR does.

In addition, whilst we are rebuilding these recipes we can also simplify the recipes for WASM command modules (`coreutils` and `grep`) by removing the `-sINVOKE_RUN=0` flag and removing `callMain` from the list of exported runtime methods. This is because we no longer need to load the JS/WASM module in one step and then call it using `callMain` as we can load and call it automatically as part of the load.

These changes could also be attempted with more complicated modules like `lua`, but I will try that separately once this is proven to work.

(Edited to correct mistake)

### DIFF
--- a/recipes/recipes_emscripten/cockle_fs/build.sh
+++ b/recipes/recipes_emscripten/cockle_fs/build.sh
@@ -5,9 +5,8 @@ emcc fs.c -o fs.js \
     -sEXPORTED_RUNTIME_METHODS=FS,PATH,ERRNO_CODES,PROXYFS \
     -sFORCE_FILESYSTEM=1 \
     -sMODULARIZE=1 \
-    -sSINGLE_FILE=1 \
     -lproxyfs.js
 
 
 mkdir -p $PREFIX/bin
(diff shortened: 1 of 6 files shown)


DerThorsten:
> with the `-sMODULARIZE=1` flag removed, which is what this PR does.

I think you ment `sSINGLE_FILE=1` flag removed

ianthomas23:
> I think you ment `sSINGLE_FILE=1` flag removed

Yes, I did mean that, I'll edit the initial comment. And I forgot to bump the build numbers too 😮

PR #1282, opened 2024-08-19
Remove use of callMain in lua

ianthomas23:
This PR removes the use of `callMain` in the `lua` build to bring it into line with `coreutils` and `grep` so that it will be usable in `cockle`.


DerThorsten:
using #1312

PR #1305, opened 2024-09-09
ping maintainers on failing PRs

DerThorsten:
ping the maintainers when a  PR to update the versions fails or has no test


martinRenou (approved):
Thanks!

PR #1312, opened 2024-09-10
remove call main and hotfix run_modularized

DerThorsten:


ianthomas23:
@DerThorsten Thanks for this, it will be really useful.

PR #1332, opened 2024-09-24
Added recipes for curl (libcurl-static)

anutosh491:
We can have 3 outputs here (curl, libcurl and libcurl-static). I guess we would not need the first two.

### DIFF
--- a/recipes/recipes_emscripten/curl/recipe.yaml
+++ b/recipes/recipes_emscripten/curl/recipe.yaml
@@ -1,0 +1,47 @@
+context:
+  name: curl
+  version: 8.10.1
+
+source:
+  url: http://curl.haxx.se/download/${{name }}-${{ version }}.tar.bz2
+  sha256: 3763cd97aae41dcf41950d23e87ae23b2edb2ce3a5b0cf678af058c391b6ae31
+
+build:
+  number: 0
+
+outputs:
+- package:
+    name: libcurl-static
+    version: ${{ version }}
+  build:
+    files:
+      - include/curl/*
+      - lib/libcurl.a
+      - lib/pkgconfig/libcurl.pc
+      - bin/curl
+      - bin/curl-config
+  requirements:
+    build:
+      - ${{ compiler("c") }}
+      - cmake
+      - make
+      - libtool
+      - pkg-config
+    host:
+      - zlib
+      - zstd  # [unix]
+      - openssl   # [unix]
+  tests:
+  - script: test -f $PREFIX/lib/libcurl.a
+
+about:
+  homepage: http://curl.haxx.se/
+  license: curl
+  license_family: MIT
+  license_file: COPYING
+  summary: tool and library for transferring data with URL syntax
+
+extra:
+  recipe-maintainers:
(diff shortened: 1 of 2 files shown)


anutosh491:
Need to 
1) Add tests
2) look into the required configuration flags (a couple might be unncessary)

anutosh491:
I did this for supporting magics in xeus-cpp but thinking about it, having `libcurl` would help us compile `RCurl` R package (cc @IsabelParedes might be interested here)

recipes/recipes_emscripten/curl/recipe.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - cmake
+      - make
+      - libtool
+      - pkg-config
+    host:
+      - zlib
+      - zstd  # [unix]
+      - openssl   # [unix]
+  tests:
+  - script: test -f $PREFIX/lib/libcurl.a
```
IsabelParedes, line 35: There's another way to check for package contents which is a bit faster but his is optional. See https://prefix-dev.github.io/rattler-build/latest/reference/recipe_file/#check-for-package-contents

IsabelParedes (changes requested):

DerThorsten:
curl will not work in the browser.

anutosh491:
> curl will very likely not work in the browser (tempted to close this PR)

I had my doubts but then I saw curl being used in wasm at few resources

1) https://curl.se/download.html (its through wasix, so search for wasix)
2) https://github.com/wapm-packages/curl
3) https://github.com/curl/curl/issues/13018#issuecomment-2045699712

For the xeus-cpp use case, even if we can catch hold of the `include/curl.h` file it would do, so i thought of raising it.

DerThorsten:
> > curl will very likely not work in the browser (tempted to close this PR)
> 
> I had my doubts but then I saw curl being used in wasm at few resources
> 
> 1. https://curl.se/download.html (its through wasix, so search for wasix)
> 2. https://github.com/wapm-packages/curl
> 3. [Duplicate symbols when compiling for wasm with emscripten curl/curl#13018 (comment)](https://github.com/curl/curl/issues/13018#issuecomment-2045699712)
> 
> And the xeus-cpp use case was just blocked since a while so I thought of raising it and this recipe is only for `libcurl-static` (not for the other 2 components which are `curl` and `libcurl`). Let me know what you think.

So while compiling curl is possible I doubt that anything meaningful will work for **the browser**.
This can work for wasi  / wasix or other platforms when they emulate the appropriate syscalls, but in the browser you will need to use **async** APIs for downloading things (ie the fetch API  ). 
If there is no dedicated code **in the curl codebase** to handle emscripten, this **will not work** to download things

anutosh491:
Thanks for the explanation here @tharun571 why exactly do we use curl in xeus-cpp. Cause I guess the above confirms we might not be able to use it.

tharun571:
@anutosh491 I am using curl to call the LLMs in xeus-cpp. As mentioned above, fetch API might be something I will try for our project.

anutosh491:
Closing as won't work.

PR #1333, opened 2024-09-24
add recipe for tornado

ukukas:


DerThorsten (changes requested):
unfortunately the 'import' tests do not work on emscripten-forge

see https://github.com/emscripten-forge/recipes/blob/cbb095e30b47c2a5d9ea1419d4fb5acc3c974992/recipes/recipes_emscripten/astropy/recipe.yaml#L38 for how we test python libs

PR #1341, opened 2024-09-27
Add r-pkgconfig

anutosh491:
Needs to be tested, not sure if this is expected to generated a ".so" file . Required for `r-tibble`


IsabelParedes:
This one is a pure R package available as `noarch` from conda-forge. It doesn't require compilation which is why it doesn't generate an *.so* file.

IsabelParedes (changes requested):
This one is a pure R package available as `noarch` from conda-forge. It doesn't require compilation which is why it doesn't generate an *.so* file.

anutosh491:
Closing as the package won't be required.

PR #1342, opened 2024-09-27
Add r-vctrs

anutosh491:
Depends on #1303

### DIFF
--- a/recipes/recipes_emscripten/r-vctrs/recipe.yaml
+++ b/recipes/recipes_emscripten/r-vctrs/recipe.yaml
@@ -1,0 +1,63 @@
+context:
+  name: r-vctrs
+  version: 0.6.5
+  r_base_version: 4.4.1
+
+package:
+  name: ${{ name }}
+  version: ${{ version }}
+
+source:
+  url:
+    - https://cran.r-project.org/src/contrib/${{ name[2:] }}_${{ version }}.tar.gz
+    - https://cloud.r-project.org/src/contrib/${{ name[2:] }}_${{ version }}.tar.gz
+  sha256: 43167d2248fd699594044b5c8f1dbb7ed163f2d64761e08ba805b04e7ec8e402
+
+build:
+  number: 0
+  script:
+    - $R CMD INSTALL $R_ARGS --no-byte-compile .
+    - rm $PREFIX/lib/R/library/grDevices/libs/cairo.so
+    - rm $PREFIX/lib/R/library/tcltk/libs/tcltk.so
+    # The activation script of cross-r-base copies cairo and tcltk from
+    # the build environment.
+
+requirements:
+  build:
+  - cross-r-base >= ${{ r_base_version }}
+  - r-glue
+  - r-rlang
+  - ${{ compiler('c') }}
+  - ${{ compiler('cxx') }}
+  host:
+  - r-base == ${{ r_base_version }}
+  - r-cli >=3.6.3
+  - r-glue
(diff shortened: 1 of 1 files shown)


anutosh491:
Hmm i get this error 
```
 × error └─ r-lifecycle 1.0.4 | 1.0.4 | 1.0.4 | 1.0.4 would require
 × error    └─ r-base >=4.3,<4.4.0a0, for which no candidates were found.
Error:   × Failed to resolve dependencies: Cannot solve the request because of: r-
  │ lifecycle >=1.0.4 cannot be installed because there are no viable options:
  │ └─ r-lifecycle 1.0.4 | 1.0.4 | 1.0.4 | 1.0.4 would require
  │    └─ r-base >=4.3,<4.4.0a0, for which no candidates were found.
  │ 
  ╰─▶ Cannot solve the request because of: r-lifecycle >=1.0.4 cannot be
      installed because there are no viable options:
      └─ r-lifecycle 1.0.4 | 1.0.4 | 1.0.4 | 1.0.4 would require
         └─ r-base >=4.3,<4.4.0a0, for which no candidates were found.
         
```
I guess r-lifecycle is built with the latest r-base (https://github.com/conda-forge/r-lifecycle-feedstock/pull/11)

IsabelParedes:
This doesn't make sense...
```
╰─λ micromamba repoquery depends r-lifecycle
Using local repodata...

Name        Version Build           Channel     Subdir
──────────────────────────────────────────────────────────
r-base      4.4.1   h745b8e0_14     conda-forge linux-64
r-cli       3.6.3   r44h0d4f4ea_1   conda-forge linux-64
r-glue      1.7.0   r44hb1dbf0f_1   conda-forge linux-64
r-lifecycle 1.0.4   r44hc72bb7e_1   conda-forge noarch
r-rlang     1.1.4   r44ha18555a_1   conda-forge linux-64
```

recipes/recipes_emscripten/r-vctrs/recipe.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - ${{ compiler('cxx') }}
+  host:
+  - r-base == ${{ r_base_version }}
+  - r-cli >=3.6.3
+  - r-glue
+  - r-rlang
+  - r-lifecycle >=1.0.4
+  run:
+  - r-base == ${{ r_base_version }}
+  - r-cli >=3.6.3
```
IsabelParedes, line 40: Good catch. It looks like we need another recipe for `r-cli` because only v3.0.1 is noarch :(

anutosh491: Yess, I got everything to build locally. So r-cli v3.0.1 is noarch so technically we can either use r-cli <3 with r-base 4.1 or we have to build latest r-cli for ourselves with r-base 4.4.1. This was pointed out by the recipe maintainers here too(https://github.com/conda-forge/r-cli-feedstock/issues/20#issuecomment-888607052)

This is how I got everything to work locally 
1) build r-cli latest
2) Use that r-cli compatible with R 4.4.1 for r-lifecycle 
3) Then r-vctrs, r-purrr, r-tibble (dependening on r-cli and r-lifecycle) was straightforward.

anutosh491:
This should be ready, just needs to be rebased on top of #1364 and #1365

anutosh491:
Okay wasn't expecting this but worked right off the bat. Thanks for the reviews @IsabelParedes :)

PR #1344, opened 2024-09-27
Add r-crayon

anutosh491:


IsabelParedes:
This is also a pure R package available as `noarch` from conda-forge. We don't need to add it.

IsabelParedes (changes requested):
This is also a pure R package available as `noarch` from conda-forge. We don't need to add it.

anutosh491:
Closing as the package won't be required.

PR #1350, opened 2024-09-27
Add cairo as dependency for harfbuzz

anutosh491:


DerThorsten:
@anutosh491 is this still relevant or can this be closed?

anutosh491:
Arghh I was trying to help @IsabelParedes out here with the recipe (I completely forgot if I was successful 😅 )

Maybe Isabel might tell us if something is pending here.

recipes/recipes_emscripten/harfbuzz/recipe.yaml
```diff
@@ -11,31 +11,35 @@ source:
...
+  - test -f $PREFIX/lib/libharfbuzz-gobject.a
+  - test -f $PREFIX/lib/libharfbuzz-cairo.a
   - test -f $PREFIX/lib/libharfbuzz-icu.a
   - test -f $PREFIX/lib/libharfbuzz-subset.a
   - test -f $PREFIX/lib/libharfbuzz.a
   - test -f $PREFIX/include/harfbuzz/hb.h
-# TODO: test hb-view, hb-shape, and hb-subset with cairo or chafa enabled
+  - test -f $PREFIX/include/harfbuzz/hb-shape.h
+  - test -f $PREFIX/include/harfbuzz/hb-subset.h
+# TODO: test hb-view with chafa enabled
```
IsabelParedes, line 42: This TODO should probably be addressed or removed. Other than that, LGTM

recipes/recipes_emscripten/harfbuzz/recipe.yaml
```diff
@@ -11,31 +11,35 @@ source:
   sha256: c758fdce8587641b00403ee0df2cd5d30cbea7803d43c65fddd76224f7b49b88
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   build:
+  # NOTE: harfbuzz doesn't need host dependencies
```
IsabelParedes, line 18: We should be using the headers from the `host` platform and not the ones from `build`, in case these have been patched (for example in `glib`). These dependencies should also be listed under host to ensure they are available for the correct platform. `glib` has not be repackaged with the new emscripten.

IsabelParedes (changes requested):

PR #1364, opened 2024-10-04
Add r-cli

anutosh491:

PR #1365, opened 2024-10-04
Add r-lifecycle

anutosh491:


anutosh491:
Should be rebased after #1364 goes in

IsabelParedes (changes requested):
This is a pure R package and **noarch**, we shouldn't need to build it.

anutosh491:
> This is a pure R package and noarch, we shouldn't need to build it.

Yes. We shouldn't need to but I found the issue we found on `r-vctrs` on any package that is r-lifecycle dependent. But I could get all those building straightaway once I built r-lifecycle. Maybe add it for now and we can remove it later ? (I guess once a newer version is released ... technically r-lifecycle should support R 4.4.1 :| ) 
I could open an issue for the same to track this if the above makes sense ?

IsabelParedes:
Let's try rebasing r-vctrs with the new r-cli and come back to this one later if we need it.

anutosh491:
Closing as not rquired.

PR #1369, opened 2024-10-08
Added xsimd 13.0.0

anutosh491:


anutosh491:
Currently packaging all architectures and types, not sure all of those are required (I guess only `include/xsimd/arch/xsimd_wasm.hpp` and `include/xsimd/types/xsimd_wasm_register.hpp` might do?) 
Maybe @JohanMabille could confirm this for me !

JohanMabille:
Packaging "all" of them is fine, xsimd is just a pure header library, nothing is compiled. When the user builds code depending on it, xsimd will detect the current architecture and "dispatch" to the appropriate header. It's easier to package headers that won't be used rather than splitting the library

JohanMabille (approved):

anutosh491:
Thanks for the review .

PR #1375, opened 2024-10-11
Update cockle packages to use emscripten 3.1.58

ianthomas23:
Update cockle packages to use emscripten 3.1.58. The packages are `cockle_fs`, `coreutils`, `grep` and `lua`. The changes comprise cherry-picking the appropriate changes from the `main` branch onto this `emscripten-3.1.58` branch and bumping the build numbers to be one larger than the previous builds.


ianthomas23:
I've tried this out locally and it works for me. It would be good to get a second opinion that the changes are OK in case I have done something stupid.

DerThorsten:
lgtm

ianthomas23:
> lgtm

Thanks.

PR #1391, opened 2024-10-23
bump pyjs to 2.4.0

trungleduc:


trungleduc:
@DerThorsten I get the same error as in the `pyjs` repo. Do we need to fix `pyjs-code-runner`?

martinRenou:
I'm a bit annoyed to suggest this, but should we drop the tests for now?

PR #1395, opened 2024-10-25
Update freetype recipe

anutosh491:

PR #1396, opened 2024-10-25
Build Cairo with freetype

anutosh491:
Here are somethings I see

1) Building cairo locally as per the recipe on master fails
2) We need a patch that was realized by the maintainers of cairo too to compile it to emscripten (merged into master but not available in 1.18.0 so should be there till next release https://gitlab.freedesktop.org/cairo/cairo/-/issues/792)
3) I think pango expects cairo to be built with freetype and fontconfig if we want to include it as a dependency

### DIFF
--- a/recipes/recipes_emscripten/cairo/build.sh
+++ b/recipes/recipes_emscripten/cairo/build.sh
@@ -2,9 +2,15 @@
 
 set -ex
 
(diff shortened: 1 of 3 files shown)


anutosh491:
Should be rebased on #1395

anutosh491:
Rebase done, CI passes as expected :)

IsabelParedes (approved):

PR #1397, opened 2024-10-25
Update glib

anutosh491:


recipes/recipes_emscripten/glib/recipe.yaml
```diff
@@ -24,6 +24,9 @@ requirements:
   - ${{ compiler('cxx') }}
   - meson
   - packaging
+  - zlib
```
IsabelParedes, line 27: Are these packages (zlib, libffi, pcre2) needed in the build environment?

anutosh491: hmmm, once sec ... confirming !

anutosh491: No they won't be required.

PR #1409, opened 2024-11-11
Debug llvm recipe without patch

anutosh491:


anutosh491:
Okay talking about the above patch

1) It seems that while cross compiling, we need to have `llvm-tblgen` (or any xx table generator like `llvm-min-tblgen` or `clang-tblgen`) built for the host and then we need to specify the location using something like this in our build script.
https://github.com/llvm/llvm-project/blob/aa15421b9ff67f362283657f74b2fa7314c25308/llvm/docs/CMake.rst?plain=1#L832
```
...
-DLLVM_TABLEGEN=$Path/bin/llvm-tblgen
...
```

2) But for our build we aren't doing this separately. We build llvm-tblgen on the go and then use it. As we need to build it for the host and not specifically for wasm, we basically need to provide a compilers like clang/clang++.

A better way to do this is add `Emscripten.cmake` here to llvm's upstream (https://github.com/llvm/llvm-project/tree/a5d09f4ad94fab718e787fb6dce7933e7742eb1b/llvm/cmake/platforms) and use that here (https://github.com/llvm/llvm-project/blob/b91b6235dee3be69b429cace319ff39f1eadfe14/llvm/cmake/modules/CrossCompile.cmake#L17)

PR #1415, opened 2024-11-13
Remove redundant flag from patch added for llvm

anutosh491:


recipes/recipes_emscripten/llvm/patches/skip_undefined_symbols_at_link_step.patch
```diff
@@ -1,13 +1,12 @@
 diff --git a/clang/lib/Interpreter/Wasm.cpp b/clang/lib/Interpreter/Wasm.cpp
-index 1001410aa0f2..fe9222b2658c 100644
+index 1001410aa0f2..1b07a947074c 100644
 --- a/clang/lib/Interpreter/Wasm.cpp
 +++ b/clang/lib/Interpreter/Wasm.cpp
-@@ -79,6 +79,8 @@ llvm::Error WasmIncrementalExecutor::addModule(PartialTranslationUnit &PTU) {
+@@ -79,6 +79,7 @@ llvm::Error WasmIncrementalExecutor::addModule(PartialTranslationUnit &PTU) {
                                            "--experimental-pic",
                                            "--no-export-dynamic",
```
anutosh491, line 7: Questionable.

Each cell maps to a side module

1) So cell1 defines and implements sqrt
2) cell 2 tries printf(sqrt(25)) .. we want sqrt symbol from `incr_module_2.wasm` and printf symbol from `xcpp.wasm`.

If incr_module_2.wasm relies on symbols from xcpp.wasm but doesn’t need to share all its own symbols globally we can use this flag to limit exports to only those explicitly listed. However, if you need all symbols available for other dynamically loaded modules, avoid --no-export-dynamic.

I am curious to know if after the first dlopen call ( incr_module_2.wasm loaded on top of main) ... does it replace xcpp.wasm as the main module (and if there is a change in the symbol table and if everything present in incr_module_2.wasm ends up in the global scope)

PR #1452, opened 2024-12-03
ncurses: include terminfo database in package

ianthomas23:
This modifies the `ncurses` recipe to include terminal database information, in other words runtime configuration files, in the `share` directory. This is needed for the upcoming `vim` recipe which needs to use the `terminfo/x/xterm-256color` to operate correctly. Rather than just building and shipping that one file, I'm including all of the database files as I expect more will be needed at some point in the future.

The changes to the recipe are at the `configure` stage, removing the `--disable-database` flag and specifying the `--build` and `--host` platforms. The latter are needed to ensure that it knows it is cross-compiling so that it uses the build platform's tool (called `tic`) to build the `terminfo` rather than the host's.

The packages `bin` directory now includes a couple more utilities and a new `share` directory which is 7.2 MB.

As far as I can tell `ncurses` is only used in the `lua` recipe so I don't think the additions here will affect any other recipes. But it would be good to get a second opinion that these changes are OK.

### DIFF
--- a/recipes/recipes_emscripten/ncurses/build.sh
+++ b/recipes/recipes_emscripten/ncurses/build.sh
@@ -11,6 +11,8 @@ export BUILD_CFLAGS="-Wno-sWASM_BIGINT"
 export BUILD_LDFLAGS="-Wno-sWASM_BIGINT"
 
 emconfigure ./configure \
+    --build=x86_64-linux-gnu \
+    --host=wasm32-unknown-emscripten \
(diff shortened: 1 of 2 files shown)


IsabelParedes:
I originally packaged `ncurses` to use it to build `readline`, but there are no plans to build `readline` anymore so these changes should be fine.

ianthomas23:
Thanks @IsabelParedes, I will merge then.

PR #1456, opened 2024-12-09
Add tree

ianthomas23:
Add new package `tree` for use in [cockle](https://github.com/jupyterlite/cockle).

It is based on the `conda-forge` recipe, here adding to the `emscripten-3.1.58` branch like all other `cockle` commands. This is a good command to have as it is easy to build and output unicode characters which are not currently supported by `cockle` commands but will be shortly.

I have not used the `-sENVIRONMENT=web.worker` emscripten flag as we need the generated javascript/wasm to support `node` to be able to run the `run_modularized` test. The default value for `ENVIRONMENT` includes `node` so in future we may as well just use the default for `cockle` commands.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -12,7 +12,7 @@ platforms = ["osx-arm64", "osx-64", "linux-64"]
 ############################################
 [feature.feature_rattler_build]
 [feature.feature_rattler_build.dependencies]
-rattler-build = ">= 0.18.2"
+rattler-build = ">=0.18.2"
 python = "3.11.*"
 typer = "*"
 curl = "*"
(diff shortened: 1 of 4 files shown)


pixi.toml
```diff
@@ -12,7 +12,7 @@ platforms = ["osx-arm64", "osx-64", "linux-64"]
 ############################################
 [feature.feature_rattler_build]
 [feature.feature_rattler_build.dependencies]
-rattler-build = ">= 0.18.2"
+rattler-build = ">=0.18.2"
```
ianthomas23, line 15: Without this I get a build-time error.

PR #1460, opened 2024-12-12
Update R packages

IsabelParedes:


IsabelParedes:
There flags are *leaking* from `r-base`, possibly the Makeconf file. To be fixed in another rebuild of `r-base`

```sh
-fvisibility=hidden -fPIC -sSIDE_MODULE=1 -g -O2
```

PR #1512, opened 2025-01-08
Update with pycrdt v0.12.2

davidbrochart:


davidbrochart:
The error seems to be:
```
Running build script
Setting up cross-python
ERROR: Version mismatch: host=3.11, build=3.12
```

IsabelParedes:
It seems to be installing 

> cross-python_emscripten-wasm32-3.11.3-py<mark>312</mark>h6dbc830_10.tar.bz2

instead of the correct

> cross-python_emscripten-wasm32-3.11.3-py<mark>311</mark>hfc8b64e_10.tar.bz2

Pinning python should fix it

```yaml
requirements:
  build:
  - cross-python_${{target_platform}}
  - python == 3.11
```

But should this package ([cross-python_emscripten-wasm32-3.11.3-py312h6dbc830_10.tar.bz2](https://beta.mamba.pm/channels/emscripten-forge/packages/cross-python_emscripten-wasm32)) with mismatched python versions even exist? @DerThorsten

davidbrochart:
pydantic-core doesn't seem to build either: https://github.com/emscripten-forge/recipes/pull/1517.

IsabelParedes:
> pydantic-core doesn't seem to build either: #1517.

This one fails because the recipe doesn't include a compiler in the build environment, and without it it installs the newer version of emscripten. 

> emscripten_emscripten-wasm32   ┆ 3.1.73 

And this package currently has some issues.

davidbrochart:
Now the error looks like a Rust compiler issue:
```
error[E0599]: no function or associated item named `default` found for struct `yrs::undo::Options` in the current scope
    --> src/undo.rs:26:36
   |
26 |         let mut options = Options::default();
   |                                    ^^^^^^^ function or associated item not found in `Options`
```
I see the version of `rustc` is 1.79.0, could it be upgraded to a newer version?

davidbrochart:
I opened #1518.

DerThorsten:
@davidbrochart 

```
#[cfg(not(target_family = "wasm"))]
impl Default for Options {
    fn default() -> Self {
        Options {
            capture_timeout_millis: 500,
            tracked_origins: HashSet::new(),
            capture_transaction: None,
            timestamp: Arc::new(crate::sync::time::SystemClock),
        }
    }
}
```
https://github.com/y-crdt/y-crdt/blob/main/yrs/src/undo.rs#L882

seems like this is not available when using wasm

davidbrochart:
I opened https://github.com/y-crdt/y-crdt/issues/520.

davidbrochart:
The error seems to be: `ModuleNotFoundError: No module named '_ssl'`.

PR #1513, opened 2025-01-08
Fix some missing licenses information

martinRenou:


IsabelParedes:
This would probably be a good time to remove all the `recipe_legacy.yaml` files

martinRenou:
Sounds good, I can look it up.

This PR merely adds license info, I guess it would be required to also put the LICENSE file in the output package.

PR #1518, opened 2025-01-09
Upgrade rust v1.83.0

davidbrochart:


DerThorsten (changes requested):
for updating rust you probably need to change the nightly dates
https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes/rust/activate-rust.sh#L14
https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes/rust/build.sh#L13

because that version here https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes/rust/recipe.yaml#L3 is not realy used

DerThorsten (approved):

DerThorsten:
thanks!

PR #1525, opened 2025-01-12
Add packages needed for ROS2WASM

Tobias-Fischer:
Hi @DerThorsten - I'm adding packages here that are needed for [ROS2WASM](https://www.arxiv.org/abs/2409.09941). It would be amazing if we could add them here so it's easier to build downstream packages without adding custom channels.


DerThorsten:
@Tobias-Fischer we can merge this (after fixing the tiny conflict).
Btw I prefer single recipes PRs, since merging many recipes at once because its a bit more error-prone.
(sometimes a single package fails because some download times out or similar, then the upload of all other pages will fail)

Tobias-Fischer:
Thanks @DerThorsten - let me actually create a new PR aiming at the 3.1.73 dev branch.

Tobias-Fischer:
Hi @DerThorsten - it would be great if you could take another look. We need to stay on the main branch for now as I get segmentation fault with the 3.1.73 branch (see prefix.dev discord)

recipes/recipes/cross-python_emscripten-wasm32/activate-cross-python.sh
DerThorsten, line 1: could you  explain why we need that change? 
Is this needed for this PR?

DerThorsten (commented):
are the cross-python changes relevant?

Tobias-Fischer:
See https://github.com/conda-forge/cross-python-feedstock/issues/24 and https://github.com/conda-forge/cross-python-feedstock/pull/69 for relevant conda-forge issues. I'd happy to defer this change to another PR if you prefer, just let me know :).

DerThorsten:
> See [conda-forge/cross-python-feedstock#24](https://github.com/conda-forge/cross-python-feedstock/issues/24) and [conda-forge/cross-python-feedstock#69](https://github.com/conda-forge/cross-python-feedstock/pull/69) for relevant conda-forge issues. I'd happy to defer this change to another PR if you prefer, just let me know :).

Since everything arround cross-python is very fragile I would prefer that In a single PR.
We probably also should make the cross-python change in the 3.1.73 branch.

In general I try to not make too much changes on the 3.1.45/main branch and hope we soon have everything on the  3.1.73 branch because maintaining these two diverging branches is a pain.

Besides that I am happy to merge your changes

Tobias-Fischer:
@DerThorsten - I've reverted the changes to `cross-python`.

Regarding 3.1.73: I would have loved to use the branch, but I got a segfault when using that branch – I'll open an issue separately about this.

Regarding cross-python: @nmarticorena: could you please do a build with the reverted `cross-python` package, and see if/where our builds fail? If they don't fail, we don't need to do anything. If they do fail, let's open a new PR re-introducing the change and detailing the issue it solves.

Tobias-Fischer:
Quick reminder @DerThorsten :)

Tobias-Fischer:
Thank you!

PR #1764, opened 2025-01-23
Fix matplotlib 3.10.0 on emscripten 3.1.73

ianthomas23:
PR #1583 claimed to pass CI but on merging the build failed. So here trying to fix it with the patch that I thought I'd already applied but perhaps didn't.


ianthomas23:
This builds fine but fails when testing both in CI and locally with
```
```

ianthomas23:
`pyee` is a dependency of `playwright-1.49.1-py313_0`. From its `conda-meta` in the local test env:

```
  "depends": [
    "greenlet 3.1.1",
    "pyee 12.0.0",
    "python >=3.13,<3.14.0a0",
    "python_abi 3.13.* *_cp313"
  ],
```

DerThorsten:
> PR #1583 claimed to pass CI but on merging the build failed. So here trying to fix it with the patch that I thought I'd already applied but perhaps didn't.

I found the reason https://github.com/emscripten-forge/recipes/pull/1787

DerThorsten:
@ianthomas23 I am making a new pyjs / pyjs-code-runner releases to convert the cryptic "error while evaluating main file: 49729112" into a usefull exception.

The large number is actually the address of an exception-ptr

ianthomas23:
Superseded by #1816.

PR #1767, opened 2025-01-23
Update r-base

IsabelParedes:
- Add patch to set `cairo` as the default bitmap type instead of using X11
- Remove `-pthread` patch which is no longer needed


recipes/recipes_emscripten/r-base/recipe.yaml
```diff
@@ -23,10 +23,10 @@ source:
     - patches/0011-Use-cross_libraries-script-when-cross-compiling.patch
     - patches/0012-Install-wasm-files.patch
     - patches/0013-Remove-png-lib-from-bitmap-libs.patch
-    - patches/0014-Build-cairo-without-pthread.patch
+    - patches/0015-Set-cairo-as-default-bitmap-type.patch
```
SylvainCorlay, line 26:
```suggestion
    - patches/0014-Set-cairo-as-default-bitmap-type.patch
```

PR #1847, opened 2025-02-03
Add Sparrow recipe [3.1.73]

anutosh491:


recipes/recipes_emscripten/sparrow/recipe.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+requirements:
+  build:
+  - ${{ compiler("cxx") }}
+  - cmake
+  - make    # [unix]
+  host:
+  - howardhinnant_date
+  run:
+  - howardhinnant_date
```
anutosh491, line 23: Guessing we need the date library at runtime too !

JohanMabille: It is not required, the date library has [run_exports](https://github.com/conda-forge/howardhinnant_date-feedstock/blob/main/recipe/meta.yaml#L19-L20)

anutosh491: Shall take care in the next update!
Thanks for the comment

PR #1859, opened 2025-02-04
Build pugixml with emsdk 3.1.73

mcbarton:
This PR is needed so I can progress with a PR in xeus-cpp https://github.com/compiler-research/xeus-cpp/pull/252/checks#step:5:35 . pugixml on the main channel is not available with emsdk=3.1.73


anutosh491:
Looking into this (might not be required I think)

anutosh491:
Closing as not required

PR #1865, opened 2025-02-06
Add xeus-javascript [3.1.73]

anutosh491:


anutosh491:
Fails I am guessing, something changed during the migration. 
So we make these changes to xeus-javascriptn, followed by a release and then update it here ?

DerThorsten:
@anutosh491 for xeus-javascript we use internal emscripten code to generate bindings for nlohman-json.
The internals changed so we need to update https://github.com/jupyter-xeus/xeus-javascript/blob/main/include/xeus-javascript/convert_json.hpp for emscripten 3.1.73

anutosh491:
> The internals changed so we need to update https://github.com/jupyter-xeus/xeus-javascript/blob/main/include/xeus-javascript/convert_json.hpp for emscripten 3.1.73

Thanks. I think a xeus 6 release is planned in the coming week. Once we have, I think we would have a good reason to make that update along with fixing this change and having a release for xeus-javascript then. I'll address this then !

PR #1898, opened 2025-02-17
m4 draft 3.1.73

DerThorsten:


wx4stg:
I've submitted a PR onto your fork that makes it work (I think). DerThorsten/recipes#15

It's a bit of a jank patch but its the only way I could get it to work.

PR #1946, opened 2025-02-28
Create typescript definition file for cockle_fs

ianthomas23:
Create typescript definition file for `cockle_fs`. This will be used in `cockle` to correctly check types of the loaded wasm module rather than use `any` for everything.


recipes/recipes_emscripten/cockle_fs/build.sh
```diff
@@ -1,10 +1,15 @@
+# Need tsc installed and available in PATH to use --emit-tsd
+npm install typescript
+export PATH=$SRC_DIR/node_modules/typescript/bin:$PATH
```
ianthomas23, line 3: @DerThorsten Are you happy with this manually modifying the `PATH`?

PR #1896, opened 2025-02-17
netcdf4 and libnetcdf

wx4stg:
Useful for meteorological and geographical datasets. I'm personally interested in having this built as I want to use xeus python jupyterlite for my library's documentation page. 

`netcdf4` is maintained by NSF Unidata. The [python bindings in `netcdf4`](https://github.com/Unidata/netcdf4-python) depend on the [underlying C library, `libnetcdf`](https://github.com/Unidata/netcdf-c).

I'm going to try my best with this... but realistically I'll probably need some help with this as I've never built anything for wasm before.

There's quite a few more deps in the tree which haven't been built for emscripten forge yet (most terrifyingly, python bindings for Open MPI). I have no idea if its mandatory to build all of them. Hopefully not.
<details>
<summary> repoquery depends netcdf4 </summary>

```

$ micromamba repoquery depends -t netcdf4 --remote
Getting repodata from channels...

conda-forge/linux-64                                        Using cache
conda-forge/noarch                                          Using cache
netcdf4[1.7.2]
  ├─ certifi[2025.1.31]
  │  └─ python[3.13.2]
  │     ├─ pip[25.0.1]
  │     ├─ tzdata[2025a]
  │     ├─ __glibc >=2.17,<3.0.a0 >>> NOT FOUND <<<
  │     ├─ libgcc[14.2.0]
  │     │  ├─ _openmp_mutex[4.5]
  │     │  │  ├─ _libgcc_mutex[0.1]
  │     │  │  └─ libgomp[14.2.0]
  │     │  │     └─ _libgcc_mutex already visited
  │     │  └─ _libgcc_mutex already visited
  │     ├─ bzip2[1.0.8]
  │     │  ├─ __glibc >=2.17,<3.0.a0 >>> NOT FOUND <<< already visited
  │     │  └─ libgcc-ng[14.2.0]
  │     │     └─ libgcc already visited
  │     ├─ libzlib[1.3.1]
  │     │  └─ libgcc-ng already visited
  │     ├─ readline[8.2]
  │     │  ├─ libgcc-ng already visited
  │     │  └─ ncurses[6.5]
  │     │     ├─ __glibc >=2.17,<3.0.a0 >>> NOT FOUND <<< already visited
  │     │     └─ libgcc already visited
  │     ├─ ncurses already visited
  │     ├─ libuuid[2.38.1]
  │     │  └─ libgcc-ng already visited
  │     ├─ libffi[3.4.6]
[...]

PR #1972, opened 2025-03-01
add netcdf4 and libnetcdf

wx4stg:
from the original PR , #1896:

> Useful for meteorological and geographical datasets. I'm personally interested in having this built as I want to use xeus python jupyterlite for my library's documentation page.

> netcdf4 is maintained by NSF Unidata. The [python bindings in netcdf4](https://github.com/Unidata/netcdf4-python) depend on the [underlying C library, libnetcdf](https://github.com/Unidata/netcdf-c).

> I'm going to try my best with this... but realistically I'll probably need some help with this as I've never built anything for wasm before.

> There's quite a few more deps in the tree which haven't been built for emscripten forge yet (most terrifyingly, python bindings for Open MPI). I have no idea if its mandatory to build all of them. Hopefully not.

### DIFF
--- a/recipes/recipes_emscripten/libnetcdf/build.sh
+++ b/recipes/recipes_emscripten/libnetcdf/build.sh
@@ -1,2 +1,2 @@
-emconfigure ./configure
+emconfigure ./configure --host=${HOST}
 emmake make
--- a/recipes/recipes_emscripten/libnetcdf/build.sh
+++ b/recipes/recipes_emscripten/libnetcdf/build.sh
@@ -1,2 +1,2 @@
-emconfigure ./configure --host=none --disable-parallel4
+emconfigure ./configure --host=none --disable-parallel4 --disable-dap --disable-dap-remote-tests
 emmake make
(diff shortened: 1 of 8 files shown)


wx4stg:
the current roadblock is that pixi is searching for only hdf5==1.12.2.* while the version on the dev prefix channel is 1.12.3. Manually specifying `hdf5==1.12.3` in the host section fixes this, but that "feels wrong", why doesn't it just find the latest..? That'll need to be fixed

But after specifying the hdf5 version, compiling fails because it tries to include "mpi.h". I'll talk to the netcdf developers to see if its possible (or, proper) to compile with mpi support disabled.

DerThorsten:
There might be a "pinning" in the variant.yaml.
Feel free to adjust the hdf5 version there.
The variant.yaml is often a bit outdated

wx4stg:
thanks!

there's the MPI issue I mentioned. I'll contact Unidata and see if its possible to build without MPI and if that's "okay to disable" for the python library. if not, I'll have to build that too, which will not be fun.

wx4stg:
mildly surprising to me that libcurl hasn't been needed by any other packages but I guess that's next..

wx4stg:
I was apparently completely wrong and m4 is just a thing that takes "m4 code" and translates it to C -- it is not actually linked against in any way.. if I had just known the right "disable" flags this could've been done in a few hours instead of a few weeks :/

PR #1977, opened 2025-03-02
r-tinyplot recipe

grantmcdermott:
Hi all,

I saw the announcement post for the new R-lite distribution. Very cool.

This PR contributes a recipe for [tinyplot](https://grantmcdermott.com/tinyplot/), a lightweight extension of the base R graphics system. (I am the maintainer.) The package is written in pure R and depends only on the base R distribution, so hopefully the build should be straightforward. 

P.S I assume the `sha256` entry on line 14 refers to the local tar.gz checksum? Let me know if I made mistake here.


IsabelParedes:
Hi @grantmcdermott, thank you for your contribution!
Since this is a pure R package and doesn't need any compilation, it will not have a shared library and that's why the test is currently failing. This package is already available on conda-forge ([tinyplot recipe](https://github.com/conda-forge/r-tinyplot-feedstock/blob/main/recipe/meta.yaml)) and can be used as it, so it's not strictly necessary to add it to emscripten-forge unless you really want to.

Hope this helps! :)

IsabelParedes:
Hi @grantmcdermott !

Installing packages on a running lite notebook doesn't work yet 😓 But we're working on a solution and it will be possible soon.
For now, the only way to install a new package is to add it to the environment and then rebuild JupyterLite.

grantmcdermott:
> Installing packages on a running lite notebook doesn't work yet 😓 But we're working on a solution and it will be possible soon.
> For now, the only way to install a new package is to add it to the environment and then rebuild JupyterLite.

Ah, makes sense. Thanks!

PR #2028, opened 2025-03-12
Update llvm to 20.1.0

anutosh491:


anutosh491:
Not sure why I should be seeing this error 
```
 │ │ In file included from $SRC_DIR/llvm/lib/Frontend/OpenACC/ACC.cpp:10:
 │ │ $SRC_DIR/build/include/llvm/Frontend/OpenACC/ACC.h.inc:192:1: error: unknown type name 'LLVM_ABI'
 │ │   192 | LLVM_ABI Directive getOpenACCDirectiveKind(llvm::StringRef Str);
 │ │       | ^
 │ │ $SRC_DIR/build/include/llvm/Frontend/OpenACC/ACC.h.inc:192:19: error: expected ';' after top level declarator
 │ │   192 | LLVM_ABI Directive getOpenACCDirectiveKind(llvm::StringRef Str);
 ```
 
 When I build it locally I see the following in build/include/llvm/Frontend/OpenAcc/Acc.h.inc
 ```
 #ifndef LLVM_OpenACC_INC
#define LLVM_OpenACC_INC

#include "llvm/ADT/ArrayRef.h"
#include "llvm/ADT/BitmaskEnum.h"
#include "llvm/Support/Compiler.h"
....
```

Which means we are including `Compiler.h` that defines `LLVM_ABI`

PR #2039, opened 2025-03-17
enable UI on openssl

wx4stg:
for some reason, building libcurl requires the UI components of openSSL to be built:

```
 │ │ wasm-ld: error: ../lib/.libs/libcurl.a(libcurl_la-openssl.o): undefined symbol: UI_OpenSSL
 │ │ wasm-ld: error: ../lib/.libs/libcurl.a(libcurl_la-openssl.o): undefined symbol: UI_OpenSSL
 │ │ wasm-ld: error: ../lib/.libs/libcurl.a(libcurl_la-openssl.o): undefined symbol: UI_OpenSSL
 │ │ wasm-ld: error: ../lib/.libs/libcurl.a(libcurl_la-openssl.o): undefined symbol: UI_OpenSSL
```

this PR removes the flag to disable UI features. I'll submit another to build curl once this is approved and pixi stops trying to pull the non-UI openSSL (and maybe some other fixes).

### DIFF
--- a/recipes/recipes_emscripten/openssl/build.sh
+++ b/recipes/recipes_emscripten/openssl/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-emconfigure ./Configure gcc -no-ui-console -DHAVE_FORK=0 -DOPENSSL_NO_SECURE_MEMORY -DNO_SYSLOG -fPIC -sWASM_BIGINT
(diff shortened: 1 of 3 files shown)


DerThorsten:
@wx4stg curl is impossible to build for wasm -- It will likely never work (because of the sandboxing of wasm).

Btw pixi is not trying anything here...the build tool is called "ratter-build". Pixi is just an "outer wrapper"

wx4stg:
gotcha--yeah unsurprisingly I had only attempted to build it with UI enabled because I needed libcurl for netcdf, except that I don't actually think that's strictly required, so if libcurl is impossible to build then I don't know of any applications outside of that.

PR #2077, opened 2025-03-25
Update patches on emsdk for 3.1.73

anutosh491:
pyodide shift to version 3.1.73 lately and updated the patches as such 

https://github.com/pyodide/pyodide/pull/5403/files

cc @DerThorsten the patch we were talking about yesterday was already a part of our source code, just didn't look updated enough, hence I guess you had to introduce the try-catch block as such 

```
                      if (!(prop in stubs)) {
                          var resolved;
                          stubs[prop] = (...args) => {
                              resolved ||= resolveSymbol(prop);
                              try{
                                  return resolved(...args)
                              }
                              catch(e){
                                  console.log(e)
                                  console.log(libName,prop)
                                  console.log("resolved", resolved)
                                  console.log("args", args)
                              }
                          }
                      }
                      return stubs[prop]
                      
```

pyodide had that patch updated and I'm updating our patches accordingly.


DerThorsten:
@anutosh491 the build numbers need to be bumped

DerThorsten:
actually this is the wrong place (and we should delete that folder).
The patch needs to go here https://github.com/emscripten-forge/recipes/tree/main/recipes/recipes/emscripten_emscripten-wasm32/patches

anutosh491:
> actually this is the wrong place (and we should delete that folder).

Argh okay. I thought we keep the patches in the emsdk folder. Thanks for pointing this out !

So the whole folder (https://github.com/emscripten-forge/recipes/tree/main/emsdk) can be removed I guess ?

DerThorsten:
> So the whole folder (https://github.com/emscripten-forge/recipes/tree/main/emsdk) can be removed I guess ?

Yes!

DerThorsten (approved):

PR #2121, opened 2025-03-27
Added osqp 1.0.0

vineetbansal:
Added recipe for [OSQP](https://osqp.org).


DerThorsten:
Thanks for the recipe!
A test in the recipe would be nice to check if the expected files are installed

vineetbansal:
Thanks @DerThorsten - I've added a couple of file tests, similar to what I see in other recipes. Let me know if anything else is needed.

PR #2150, opened 2025-04-02
Create lib dir if not existant

AntoinePrv:


IsabelParedes (approved):

IsabelParedes (commented):
We should bump the build number too before merging

PR #2201, opened 2025-04-07
Add logos

IsabelParedes:
Added several logo files in case they are needed in the other repos.

image
image


DerThorsten:
nice! Thanks!

DerThorsten (approved):

PR #2270, opened 2025-04-23
Bump xeus-cpp build number to use latest xeus-lite

ianthomas23:
Bump `xeus-cpp` build number to rebuild using latest `xeus-lite 3.1.0` that includes fixes for reading from `stdin`.


anutosh491:
Hey Ian, 

Thanks a lot for this. Trying it out soon :)

ianthomas23:
It needs changes in both `jupyterlite/xeus` and `jupyterlite/jupyterlite` before it is possible to try this out.

anutosh491:
Ahh okay so this one just handles the xeus-lite side of things ? Okay that makes sense

ianthomas23:
Yes, this will now ask for `stdin` in the correct way but JupyterLite will not yet respond in the correct way.

anutosh491 (approved):
LGTM

PR #2287, opened 2025-04-30
Update llvm to 20.1.4

anutosh491:


anutosh491:
One of my backport PRs on wasm-ld was not included in the release https://github.com/llvm/llvm-project/pull/137620 (probably because I had to manually backport it due to some merge commits and the release maintainer probably missed it)

But it is critical though and hence I've added it as a patch right now which I shall remove in 20.1.5 which should be out in a couple week or so !

SylvainCorlay (approved):

SylvainCorlay:
I let @DerThorsten merge!

PR #2293, opened 2025-04-30
Update [de]activation scripts for consistency

IsabelParedes:
This standardizes the naming of the activation/deactivation scripts so that they run in the correct order.
- The activation scripts of the compiler packages (emscripen, rust, flang) should run first, so they are renamed to `activation_<pkg>.sh`.
- The activation scripts of the cross packages (cross-python, cross-r) should run last, so they are renamed to `activation_z-<pkg>.sh`

Changes to `cross-python` will come in a separate PR, but it already has the correct names.


IsabelParedes:
Tested with `pillow`, `pycrdt` and `regex`

DerThorsten (approved):

PR #2307, opened 2025-05-07
Add a shared build for xwidgets

anutosh491:
This PR introduces a shared build for xwidgets so that we can load it during runtime while using it with xeus-cpp-lite and can avoid any manual linking/loading.

I have named it as xwidgets_shared to not disturb the static archives produced by the xwidgets recipe itself. This introduces a patch (that shall be upstreamed in the near future)


anutosh491:
Currently I end up here (this recipe or the shared build doesn't look like the culprit here to me)

image

I have been able to get it working in the past using libxwidgets.a (through a manual linking with target xcpp)

image

SylvainCorlay:
This looks good to me.

SylvainCorlay (approved):

SylvainCorlay:
This should probably make use of the shared xeus library now.

SylvainCorlay (changes requested):
Let's link with xeus shared.

anutosh491:
Superseded by #2402

PR #2327, opened 2025-05-17
Add `rustworkx` alpha

IvanIsCoding:
Related: https://github.com/pyodide/pyodide-recipes/pull/90 and https://github.com/Qiskit/rustworkx/pull/1447

In short, this adds a very experimental rustworkx to emscripten-forge. It works with Pyodide, so I confirmed it worked with emscripten-forge as well: the tests passed after running `pixi run build-emscripten-wasm32-pkg recipes/recipes_emscripten/rustworkx`

At the current status, the other rustworkx author has not yet merged the Pyodide/emscripten-forge support. It also might take a while for us to publish 0.17.0. So I called it a pre-release with 0.17.0a3 and hosted it in my GitHub fork's release. If that is not ok, let me know.

### DIFF
--- a/recipes/recipes_emscripten/rustworkx/recipe.yaml
+++ b/recipes/recipes_emscripten/rustworkx/recipe.yaml
@@ -16,6 +16,7 @@ build:
 requirements:
   build:
   - cross-python_${{target_platform}}
+  - ${{ compiler("c") }}
   - cffi
   - setuptools-rust
   - rust
--- a/recipes/recipes_emscripten/rustworkx/recipe.yaml
+++ b/recipes/recipes_emscripten/rustworkx/recipe.yaml
@@ -1,0 +1,41 @@
+context:
+  name: rustworkx
+  version: 0.17.0a3
+
+package:
+  name: ${{name}}
+  version: ${{ version }}
+
+source:
+- url: https://github.com/IvanIsCoding/rustworkx/releases/download/v0.17.0a3/rustworkx-0.17.0a3.tar.gz
+  sha256: 8bd0c295134e2b0c03808d4e69428b41153849db6488839084a78793c337f191
+
+build:
+  number: 0
+
+requirements:
+  build:
+  - cross-python_${{target_platform}}
+  - cffi
+  - setuptools-rust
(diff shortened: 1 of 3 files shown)


recipes/recipes_emscripten/rustworkx/recipe.yaml
```diff
@@ -0,0 +1,41 @@
+context:
+  name: rustworkx
+  version: 0.17.0a3
+
+package:
+  name: ${{name}}
+  version: ${{ version }}
+
+source:
+- url: https://github.com/IvanIsCoding/rustworkx/releases/download/v0.17.0a3/rustworkx-0.17.0a3.tar.gz
```
IvanIsCoding, line 10: Again, this is one of the few shortfalls of this PR. I hope the forked alpha is OK, we never tried a pre-release on `rustworkx`'s side

recipes/recipes_emscripten/rustworkx/recipe.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - rust
+  - rust-src
+  - maturin
+
+  host:
+  - python
+  - openssl
+  - cffi
+  run:
+  - cffi
```
IvanIsCoding, line 30: I copied this from the other Rust recipes. I am not sure it is necessary, but the tests passed so I am not going to touch it.

DerThorsten:
Thanks for the recipe

recipes/recipes_emscripten/rustworkx/recipe.yaml
DerThorsten: The compiler package should be added to the build dependencies, otherwise we might not get the emscripten abi package

DerThorsten (changes requested):

IvanIsCoding:
I added the C compiler based on the recipe for `conda-forge`: https://github.com/conda-forge/rustworkx-feedstock/blob/4635c4cba4232fa7e9d2bcbd035e6d978dc998d3/recipe/recipe.yaml#L30.

DerThorsten (approved):

PR #2334, opened 2025-05-20
Add nlohmann_json to xeus-cpp as runtime dependecy

anutosh491:
Needed for running rich rendering and other examples with xeus-cpp-lite in notebook link.


SylvainCorlay:
The pin in host is due to the fact that xeus core depends on nlohmann_json ?

SylvainCorlay:
Alright. LGTM.

SylvainCorlay (approved):

PR #2338, opened 2025-05-21
Add recipe for clang's resource dir

anutosh491:


anutosh491:
Hi, 

Currently cppinterop doesn't extend llvm's clang resource dir to xeus-cpp.

So though we technically don't depend on llvm. This recipe just monitors the "lib/clang/20/include" directory from the llvm recipe that we need to put to use.

This will lie in the host env and shall be preloaded from here https://github.com/compiler-research/xeus-cpp/blob/main/CMakeLists.txt#L441

so that xeus-cpp-lite can have access to wasm_simd128.h required to run xsimd !

anutosh491:
> This will lie in the host env and shall be preloaded from here https://github.com/compiler-research/xeus-cpp/blob/main/CMakeLists.txt#L441

Actually let's not do this but as it as a runtime dependency for cppinterop on emscripten-forge (cause technically cppinterop should be responsible for this)

So we change nothing : cppinterop is just cppinterop + these headers.
once we do that, let me add the -msimd128 flag to all kernels by default !

SylvainCorlay (approved):

PR #2339, opened 2025-05-21
[wip] Removed cross-compile patch from llvm-emscripten build

kr-2003:
EDIT by Anutosh : For anyone interested in this change. Serge and I had introduced a hacky patch to build llvm back in Dec 2023. We knew it could be addressed sometime down the line, just had find time for it. Here's the context and references to support this change https://github.com/compiler-research/CppInterOp/pull/583#issuecomment-2896618086


recipes/recipes_emscripten/llvm/build.sh
```diff
@@ -9,28 +16,27 @@ export LDFLAGS=""
...
+    -DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4 -fexceptions" \
+    -DLLVM_INCLUDE_BENCHMARKS=OFF \
+    -DLLVM_INCLUDE_EXAMPLES=OFF \
+    -DLLVM_INCLUDE_TESTS=OFF \
+    -DLLVM_ENABLE_THREADS=OFF \
+    -DLLVM_BUILD_TOOLS=OFF \
+    -DLLVM_ENABLE_LIBPFM=OFF \
+    -DCLANG_BUILD_TOOLS=OFF \
+    -DLLVM_TABLEGEN=$NATIVE_DIR/llvm-tblgen \
+    -DCLANG_TABLEGEN=$NATIVE_DIR/clang-tblgen \
```
anutosh491, line 39: I see only the last 2 configs being extra. Can we not change the other configs if possible ?

recipes/recipes_emscripten/llvm/recipe.yaml
```diff
@@ -10,7 +10,6 @@ source:
     }}.tar.gz
   sha256: 205c436f93a7e185cf9d63ae57e645f23561a4d1bf854f2c5bd78995a43a5c05
   patches:
-  - patches/cross_compile.patch
   - patches/shift_temporary_files_to_tmp_dir.patch
   - patches/enable_exception_handling.patch
```
anutosh491, line 15: On line 17, please update the build number to 1

anutosh491:
Looks like the error occurs because emscripten-forge might be forcing the toolchain file, but we're building llvm-tblgen etc for the host correct ? This means we need to suppress it for 1st half and later we can restore the toolchain

kr-2003:
> Looks like the error occurs because emscripten-forge might be forcing the toolchain file, but we're building llvm-tblgen etc for the host correct ? This means we need to suppress it for 1st half and later we can restore the toolchain

By supress do you mean "unsetting the CMAKE_TOOLCHAIN_FILE" while building for host?

anutosh491:
> By supress do you mean "unsetting the CMAKE_TOOLCHAIN_FILE" while building for host?

Can possibly trying that (passing empty string for the host)

anutosh491:
One small thing, I see this while building the llvm-tblgen
```
cmake -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=host -DCMAKE_BUILD_TYPE=Release ../llvm/ -DCMAKE_CROSSCOMPILING_EMULATOR=$BUILD_PREFIX/bin/node
```
I don't think we need the cross compiling emulator as we're building for host. Could you address this too ?

kr-2003:
> One small thing, I see this while building the llvm-tblgen
> 
> ```
> cmake -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=host -DCMAKE_BUILD_TYPE=Release ../llvm/ -DCMAKE_CROSSCOMPILING_EMULATOR=$BUILD_PREFIX/bin/node
> ```
> 
> I don't think we need the cross compiling emulator as we're building for host. Could you address this too ?

Done

anutosh491:
Thanks for the changes, here hopefully everything build fine.

recipes/recipes_emscripten/llvm/build.sh
```diff
@@ -1,8 +1,26 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DLLVM_ENABLE_PROJECTS=clang \
+    -DLLVM_TARGETS_TO_BUILD=host \
+    -DCMAKE_TOOLCHAIN_FILE="" \
+    -DCMAKE_CROSSCOMPILING_EMULATOR=""
+  make llvm-tblgen clang-tblgen -j$(nproc --all)
+  export NATIVE_TOOLS_DIR=$PWD/bin
+  cd ..
+)
+
```
anutosh491, line 16: Here we want to build llvm-tblgen and clang-tblgen for the host (no cross compilation here)

If emscripten-forge has some configs not allowing this. Probably in the build section we can get llvm-tblgen already built from conda-forge and then just point to it below I geuss !

recipes/recipes_emscripten/llvm/build.sh
```diff
@@ -30,7 +48,10 @@ emcmake cmake ${CMAKE_ARGS} -S ../llvm -B .         \
     -DCLANG_ENABLE_ARCMT=OFF                        \
     -DCLANG_ENABLE_BOOTSTRAP=OFF                    \
     -DCLANG_BUILD_TOOLS=OFF                         \
-    -DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4 -fexceptions"
+    -DCMAKE_CXX_FLAGS="-Dwait4=__syscall_wait4 -fexceptions" \
+    -DLLVM_TABLEGEN=$NATIVE_DIR/llvm-tblgen         \
+    -DCLANG_TABLEGEN=$NATIVE_DIR/clang-tblgen       \
+    -DLLVM_ENABLE_LIBPFM=OFF                        \
```
anutosh491, line 54: Not required isn't it ?

anutosh491:
Hey @kr-2003,

Through conda-forge (or through any llvm release) ,

We won't be having direct access to clang-tblgen. Here's the reason why the llvm-tblgen binary is provided but clang-tblgen is not !

https://discourse.llvm.org/t/clang-tblgen-not-installed/52997

Which means we need to build our own llvm-tblgen and clang-tblgen !

So we need to move along the lines of what you've done 
1) we deactivate the script
2) build what we need for the build arch
3) activate the script
4) Build llvm

kr-2003:
Hey @anutosh491, 

Thanks. 

I have pushed the changes. Made two new scripts (i.e. activats.sh and deactivate.sh) inside the recipe-directory.

kr-2003:
Hey @DerThorsten,

Should each recipe have their own activation/deactivation scripts? Is it possible for us to use the scripts under ``/recipes/recipes/emscripten_emscripten-wasm32`` in ``recipes/recipes_emscripten/llvm/build.sh``.

DerThorsten:
> Hey @DerThorsten,
> 
> Should each recipe have their own activation/deactivation scripts? Is it possible for us to use the scripts under `/recipes/recipes/emscripten_emscripten-wasm32` in `recipes/recipes_emscripten/llvm/build.sh`.

I think in general, we want to avoid activation scripts when possible.
It might be possible to call the de-activation script by hand, but its a hack.

Also I am a bit unsure why we need to add new 

> Hey @kr-2003,
> 
> Through conda-forge (or through any llvm release) ,
> 
> We won't be having direct access to clang-tblgen. Here's the reason why the llvm-tblgen binary is provided but clang-tblgen is not !
> 
> https://discourse.llvm.org/t/clang-tblgen-not-installed/52997
> 
> Which means we need to build our own llvm-tblgen and clang-tblgen !
> 
> So we need to move along the lines of what you've done
> 
> 1. we deactivate the script
> 2. build what we need for the build arch
> 3. activate the script
> 4. Build llvm

if you build for the "build-arch" then there is no emscripten whatsoever.  So why would you need to do any disabling of activation scripts?

kr-2003:
> if you build for the "build-arch" then there is no emscripten whatsoever. So why would you need to do any disabling of activation scripts?

Yes, correct. My bad. I have moved ``llvm-tblgen`` and ``clang-tblgen`` to be built under ``llvm-tblgen`` pkg name in "build-arch".  And then using this build with ``llvm-empscripten`` build.

anutosh491 (approved):
Hey @kr-2003 , this looks good.

I'll allow @DerThorsten to take the final call just to confirm everything's in place.

For some context 
1) llvm's emscripten build need llvm-tblgen/clang-tblgen being built for the build arch and then being shared through `-DLLVM_TABLEGEN` while building (discussion with Serge and other references can be found here https://github.com/compiler-research/CppInterOp/pull/583#issuecomment-2896618086)

2) We were using a hacky patch to build llvm-tblgen at buildtime and then using it on the go. But would be much better without it.

3) Hence this recipe should add a build arch recipe for llvm-tblgen and update the emscripten recipe for llvm to reference the location of the tablegen executable.

DerThorsten (approved):

kr-2003:
> Hey @kr-2003,
> 
> Thanks for the changes here. Now that this has gone in, we need to test if this llvm build is working as per what we want. As a follow up
> 
> 1. Fetch latest llvm locally once hosted (should take about 2 hours), build cppinterop on top of it and run the emscripten tests. All tests should pass as expected
> 2. Once that is done, update cppinterop's recipe to use this llvm build (you would just need to increment the build number by 1)
> 
> Ping me once this is done.

Its working 🚀 .
Screenshot 2025-05-27 at 4 55 21 PM

PR #2345, opened 2025-05-22
CppInterOp 1.7.0

SylvainCorlay:
@anutosh491 this is removing the patch from the previous version, which is at least partially included in the new release.


anutosh491:
Thanks for this 

I think this is good for the zeroth build number. I have some minute changes in mind. That I shall make soon !

anutosh491 (approved):

PR #2354, opened 2025-05-27
Building xeus-cpp on top of latest CppInterOp

kr-2003:


anutosh491:
Hey Sylvain, we had recently llvm (without the hacky patch) -> then built cppinterop on top of it and then building xeus-cpp on top of it. 

Taking this opportunity to add xeus as a runtime dependency !

SylvainCorlay (approved):

PR #2357, opened 2025-05-29
Fix rich display feature for symengine

anutosh491:


recipes/recipes_emscripten/symengine/patches/namespace.patch
```diff
@@ -21,7 +12,7 @@ index 5d0c63e73..bfdc26398 100644
 -
 -#ifdef __CLING__
 +// Utility functions for xeus-cpp
-+#ifdef defined(__CLANG_REPL__)
++#if defined(__CLANG_REPL__)
```
anutosh491, line 15: image

SylvainCorlay: Excellent catch!

PR #2358, opened 2025-05-30
Add a shared build for xeus

anutosh491:
This would help xeus-cpp-lite use xwdidgets/xcanvas etc 

image

### DIFF
--- a/recipes/recipes_emscripten/xeus_shared/patches/shared.patch
+++ b/recipes/recipes_emscripten/xeus_shared/patches/shared.patch
@@ -1,0 +1,27 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index a8147d3..f3a99c6 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -59,8 +59,12 @@ option(XEUS_BUILD_TESTS "xeus test suite" OFF)
+ # ===================================
+ 
+ if(EMSCRIPTEN)
+-    SET(XEUS_BUILD_SHARED_LIBS OFF)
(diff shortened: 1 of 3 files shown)


recipes/recipes_emscripten/xeus_shared/patches/shared.patch
```diff
@@ -0,0 +1,27 @@
...
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -59,8 +59,12 @@ option(XEUS_BUILD_TESTS "xeus test suite" OFF)
+ # ===================================
+ 
+ if(EMSCRIPTEN)
+-    SET(XEUS_BUILD_SHARED_LIBS OFF)
+-    SET(XEUS_BUILD_STATIC_LIBS ON)
++    SET(XEUS_BUILD_SHARED_LIBS ON)
++    SET(XEUS_BUILD_STATIC_LIBS OFF)
```
SylvainCorlay, line 12: I think we should rather completely remove the `SET(XEUS_BUILD_SHARED_LIBS` and `SET(XEUS_BUILD_STATIC_LIBS` instructions, and pass the options command line. After all, these are command line options, and this would make the emscripten case less special.

recipes/recipes_emscripten/xeus_shared/recipe.yaml
```diff
@@ -0,0 +1,39 @@
+context:
+  version: 5.2.2
+
+package:
+  name: xeus
```
SylvainCorlay, line 5: Why do we need a different recipe from the xeus main xeus one?

Note: this sems to be naming the package `xeus` anyways.

PR #2367, opened 2025-06-03
Fix Initialization logs for xeus-cpp-lite

anutosh491:
As can be seen here https://github.com/llvm/llvm-project/blob/9801fdf669bf6e61e276732a95c2f885cb2a39d9/clang/lib/Driver/Driver.cpp#L2445 , llvm::errs is used for verbose logging, leading to the initialization logs being treated as errors 

image

### DIFF
--- a/recipes/recipes_emscripten/llvm/patches/fix_clang_keywords.patch
+++ b/recipes/recipes_emscripten/llvm/patches/fix_clang_keywords.patch
@@ -1,0 +1,20 @@
+diff --git a/clang/lib/Parse/ParseTentative.cpp b/clang/lib/Parse/ParseTentative.cpp
+index ff27ef70944a..27cdf2f90f51 100644
+--- a/clang/lib/Parse/ParseTentative.cpp
++++ b/clang/lib/Parse/ParseTentative.cpp
+@@ -1504,6 +1504,7 @@ Parser::isCXXDeclarationSpecifier(ImplicitTypenameContext AllowImplicitTypename,
+   case tok::kw_inline:
+   case tok::kw_virtual:
+   case tok::kw_explicit:
++  case tok::kw__Noreturn:
+ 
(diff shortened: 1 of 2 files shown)


anutosh491:
This patch should help us with the above, changing llvm::errs to llvm::outs resulting into 

image

recipes/recipes_emscripten/llvm/recipe.yaml
```diff
@@ -12,9 +12,12 @@ source:
   patches:
   - patches/shift_temporary_files_to_tmp_dir.patch
   - patches/enable_exception_handling.patch
+  - patches/fix_initialization_logs.patch
+  - patches/fix_transitive_headers.patch
```
anutosh491, line 16: The patch fixing including transitive headers will be removed by the next release 20.1.7 

Should help us with catching errors instead of hitting the assert block as seen

image

recipes/recipes_emscripten/llvm/patches/fix_clang_keywords.patch
```diff
@@ -0,0 +1,20 @@
...
+     // Modules
+   case tok::kw___module_private__:
+@@ -1558,6 +1559,7 @@ Parser::isCXXDeclarationSpecifier(ImplicitTypenameContext AllowImplicitTypename,
+     // GNU
+   case tok::kw_restrict:
+   case tok::kw__Complex:
++  case tok::kw__Imaginary:
+   case tok::kw___attribute:
+   case tok::kw___auto_type:
+     return TPResult::True;
```
anutosh491, line 20: Upstream patch has been merged and we can get rid of this by the next release (12th June)

Fixes this issue https://github.com/compiler-research/xeus-cpp/issues/320

SylvainCorlay (approved):

PR #2379, opened 2025-06-05
Add package lz4-c

davidbrochart:


recipes/recipes_emscripten/lz4-c/recipe.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+package:
+  name: ${{ pkg_name }}
+  version: ${{ version }}
+
+source:
+  url: https://github.com/${{ name }}/${{ name }}/archive/v${{ version }}.tar.gz
+  sha256: 537512904744b35e232912055ccf8ec66d768639ff3abe5788d90d792ec5f48b
+  #patches:
+  #- patches/CVE-2021-3520.patch
```
davidbrochart, line 14: It looks like in the original feedstock the patches are not applied, I guess I should just remove this.

recipes/recipes_emscripten/lz4-c/recipe.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+build:
+  number: 1
+
+requirements:
+  build:
+    - ${{ compiler('c') }}
+    #- ${{ stdlib("c") }}
```
davidbrochart, line 22: Is is fine to remove this?

recipes/recipes_emscripten/lz4-c/recipe.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - ${{ compiler('cxx') }}
+    - make
+    - patch
+  #host:
+  #  - ${{ pin_subpackage('lz4-c', exact=True) }}
```
davidbrochart, line 27: Is this fine to remove this?

davidbrochart:
This recipe comes from [here](https://github.com/conda-forge/lz4-c-feedstock).
I'm not sure if there should only be the static library, or also the shared library?

recipes/recipes_emscripten/lz4-c/recipe.yaml
```diff
@@ -0,0 +1,52 @@
...
+  run:
+    - ${{ pin_subpackage('lz4-c', exact=True) }}
+
+tests:
+#  files:
+#    - test_static_cmake
```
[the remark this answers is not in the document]
IsabelParedes: I don't like this cmake test, and it only checks if the package is there. Instead, we can do something like the following to make sure that the important files in your package are actually there:

```yaml
tests:
  - package_contents:
      files:
        - lib/liblz4.a
        - include/lz4.h
```

I also see a few files under *bin/*, simple tests for these can be something like

```
- script:
    - node ${PREFIX}/bin/lz4 --version
  requirements:
    build:
      - nodejs
```

However, the corresponding *lz4.wasm* would also need to be included in the *bin/* directory for those tests to work.

recipes/recipes_emscripten/lz4-c/build.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -ex
+
+export CFLAGS="${CFLAGS} -O3 -fPIC"
+export LDFLAGS="${LDFLAGS} -Wl,-rpath,${PREFIX}/lib -L${PREFIX}/lib"
```
IsabelParedes, line 5: We can probably remove this line. `-rpath` is not supported and the default `LDFLAGS` should work well.

recipes/recipes_emscripten/lz4-c/build.sh
```diff
@@ -0,0 +1,32 @@
...
+  $DATAGEN -g16KB | $LZ4 -9 | $LZ4 -t
+  $DATAGEN | $LZ4 | $LZ4 -t
+  $DATAGEN -g6M -P99 | $LZ4 -9BD | $LZ4 -t
+  $DATAGEN -g17M | $LZ4 -9v | $LZ4 -qt
+  $DATAGEN -g33M | $LZ4 --no-frame-crc | $LZ4 -t
+  $DATAGEN -g256MB | $LZ4 -vqB4D | $LZ4 -t
+fi
+
+# Install
+make install PREFIX=${PREFIX}
```
IsabelParedes, line 32: Could you add a verbose option? Maybe `VERBOSE=1` or `V=1`?

recipes/recipes_emscripten/lz4-c/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -ex
+
+export CFLAGS="${CFLAGS} -O3 -fPIC"
+export LDFLAGS="${LDFLAGS} -lrt"
+
+# Build
+make -j${CPU_COUNT} PREFIX=${PREFIX} VERBOSE=1
```
davidbrochart, line 8: Not sure if it should be:
```suggestion
VERBOSE=1 make -j${CPU_COUNT} PREFIX=${PREFIX}
```

IsabelParedes: The way you have it seems to be working already

PR #2386, opened 2025-06-06
cross-python-cmake-find-python

DerThorsten:
python extensions build with cmake, in particular extensions using nanobind rely on `FindPython.cmake` in the **build-prefix** to do the right thing. But of course it does not :/.

This ment that packages had to do horrible stuff to avoid running into issues from `FindPython.cmake`.
The following are the horrors we had to do (ie create a MyFakePythonModule and and alias from `MyFakePythonModule`to `Python::Module`):
```cmake
if(EMSCRIPTEN)

    # Create a target that holds Python module information
    add_library(MyFakePythonModule INTERFACE)

    # Define the include directories
    set_target_properties(MyFakePythonModule PROPERTIES
        INTERFACE_INCLUDE_DIRECTORIES "${Python_INCLUDE_DIRS}"
    )

    # Define the link libraries (the actual Python library)
    set_target_properties(MyFakePythonModule PROPERTIES
        INTERFACE_LINK_LIBRARIES "${Python_LIBRARIES}"
    )

    # Optionally, if nanobind uses the Python executable path:
    set_target_properties(MyFakePythonModule PROPERTIES
        INTERFACE_COMPILE_DEFINITIONS "PYTHON_EXECUTABLE=\"${Python_EXECUTABLE}\""
    )

    # Attempt to create an alias (might not fully satisfy the check)
    add_library(Python::Module ALIAS MyFakePythonModule)
else()
    # Find Python
    find_package(Python COMPONENTS Interpreter Development REQUIRED)
endif()
```

To avoid this, we overwrite the build-prefix FindPython with a custom one in the cross-python activation script.
We build `pyb2d3` with the described changed to see if this works!

Not only the `CMakeLists.txt` are drastically simplified, but also the build script.
(see the diff)

### DIFF
--- a/recipes/recipes/cross-python_emscripten-wasm32/FindPython.cmake
+++ b/recipes/recipes/cross-python_emscripten-wasm32/FindPython.cmake
@@ -1,0 +1,128 @@
+# This is your custom FindPython.cmake
+
+set_property(GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS TRUE)
+set(CMAKE_SHARED_LIBRARY_CREATE_C_FLAGS "-s SIDE_MODULE=1")
+set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "-s SIDE_MODULE=1")
+set(CMAKE_STRIP FALSE)  # used by default in pybind11 on .so modules
+
+
+# Indicate that Python and its components are found
+set(Python_FOUND TRUE)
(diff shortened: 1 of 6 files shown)


DerThorsten:
Python extensions which are build with cmake (in particular nanobind) relies on FindPython.cmake.
It will try to use the FindPython.cmake from the build prefix which does not work properly.
This PR adds logic to cross-python to overwrite the FindPython.cmake in the build prefix.

PR #2398, opened 2025-06-09
Package psutil

davidbrochart:
But I'm not sure this can be compiled to WASM, since processes are a feature of the OS.


davidbrochart:
> But I'm not sure this can be compiled to WASM, since processes are a feature of the OS.

Yes:
```
psutil/_psutil_linux.c:13:10: fatal error: 'linux/ethtool.h' file not found
```

PR #2403, opened 2025-06-11
Update xcanvas to be usable with xeus-cpp-lite

anutosh491:


anutosh491:
Okay let's try this out !

PR #2405, opened 2025-06-11
Add xwidgets as a runtime dependency for xcanvas

anutosh491:
I just realize that before loading libxcanvas.so, we need to load libxwidget.so in xeus-cpp-lite which wasn't being addressed.


recipes/recipes_emscripten/xcanvas/recipe.yaml
```diff
@@ -23,6 +23,8 @@ requirements:
   - xproperty >=0.12.0,<0.13
   - nlohmann_json >=3.12
   - xwidgets >=0.29.0,<0.30
+  run:
+  - xwidgets >=0.29.0,<0.30
```
JohanMabille, line 27: We should add a run_exports section to the xwidgets recipes instead.

JohanMabille (changes requested):

PR #2411, opened 2025-06-12
Use run_exports for xwidgets

anutosh491:


recipes/recipes_emscripten/xwidgets/recipe.yaml
```diff
@@ -12,21 +12,23 @@ source:
...
+  - nlohmann_json >=3.12
   - xeus >=5,<6
   - xproperty >=0.12.0,<0.13
   run:
-  - nlohmann_json >=3.11
+  - nlohmann_json >=3.12
   - xeus >=5,<6
   - xproperty >=0.12.0,<0.13
+  run_exports:
+  - ${{ pin_subpackage("xwidgets", upper_bound="x.x") }}
```
JohanMabille, line 31: I think it should be ` max_pin`  instead of ` upper_bound`  (or mayabe I am not aware of the existenc of `upper_bound`, first time I see it).

anutosh491: Ahh I used max_pin initially (look at the commit before this) but the CI asked me to replace it with upper_bound !

JohanMabille: Oh ok so that's probably from the new recipes format.

PR #2423, opened 2025-06-13
Replicate CppInterOp's llvm build

anutosh491:
Problem:

At this point, we've tired quite some things to get the llvm build working. It is just the `nlohmann/json.hpp` header that is troubling is (by itself or while being included by xwidgets/xcanvas etc). This has lot of template influence and llvm ends up in a recursive infinite loop calling `printTemplateSpecializationBefore`

Approach:

Building llvm locally without any configs inspired by emscripten-forge or the patched emsdk works perfectly.
This is what CppInterOp is doing 
https://github.com/compiler-research/CppInterOp/blob/main/.github/workflows/emscripten.yml#L207-L228

And the link on their main is working perfectly 

image

https://compiler-research.org/CppInterOp/lab/index.html

Hence this PR tries to replicate the build there.

### DIFF
--- a/recipes/recipes_emscripten/llvm/build.sh
+++ b/recipes/recipes_emscripten/llvm/build.sh
@@ -6,6 +6,16 @@ cd build
 export CMAKE_PREFIX_PATH=$PREFIX
 export CMAKE_SYSTEM_PREFIX_PATH=$PREFIX
 
+unset EM_FORGE_OPTFLAGS
+unset EM_FORGE_DBGFLAGS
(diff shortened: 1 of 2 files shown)


anutosh491:
In our case, as can be seen here https://compiler-research.org/xeus-cpp/lab/index.html

If we forcefully break the recursive call

image

we're good, but if we don't 

image

So most headers are still fine. But headers that have some recursive template inspiration might be problematic here. And I believe the flags (optimization level especially) might be responsible for the recursion depth here !

PR #2431, opened 2025-06-16
Add orjson

anutosh491:


anutosh491:
I think this would be needed for ipycanvas (which I would like to add as a run dependency for xcanvas)
I see this 
```
Error:   × Test failed: failed to setup test environment: Cannot solve the request
  │ because of: xcanvas ==0.5.1 h2072262_4 cannot be installed because there
  │ are no viable options:
  │ └─ xcanvas 0.5.1 would require
  │    └─ ipycanvas 0.11.0.*, which cannot be installed because there are no
  │ viable options:
  │       └─ ipycanvas 0.11.0 would require
  │          └─ orjson *, for which no candidates were found.
  │ The following packages are incompatible
  │ └─ xcanvas ==0.5.1 h2072262_4 cannot be installed because there are no
  │ viable options:
  │    └─ xcanvas 0.5.1 is excluded because due to strict channel priority not
  │ using this option from: 'file:///home/runner/work/recipes/recipes/output/'
  │ 

```

anutosh491:
Orjson was also not present on the latest 3.1.73 branch 

So should be a good addition.

PR #2482, opened 2025-06-30
Add narwhals

jjerphan:
One of the dependencies of `skore`, also useful for different workflows using dataframes for non-exclusively scikit-learn et al (see https://github.com/scikit-learn/scikit-learn/issues/31049).

Also a first contribution to `emscripten-forge`.


jjerphan:
Indeed, I did not know `emscripten-forge` has (to be used with) a fallback on `conda-forge`.

PR #2600, opened 2025-07-22
Package fps-kernel-web-worker

davidbrochart:


IsabelParedes (changes requested):
The package is currently empty

```
Files in package:

  - info/about.json
  - info/hash_input.json
  - info/index.json
  - info/licenses/COPYING.md
  - info/paths.json
  - info/recipe/recipe.yaml
  - info/recipe/rendered_recipe.yaml
  - info/recipe/variant_config.yaml
  - info/tests/tests.yaml
```

davidbrochart:
@IsabelParedes I got this error:
```
Error:   × Failed to resolve dependencies: Cannot solve the request because of: No
  │ candidates were found for cross-python_emscripten-wasm32 *.
  │ 
  ╰─▶ Cannot solve the request because of: No candidates were found for cross-
      python_emscripten-wasm32 *.
```

davidbrochart:
I don't quite understand the error, does it mean that my test failed?

IsabelParedes:
> I don't quite understand the error, does it mean that my test failed?

It seems that having `pyjs` in the run dependencies causes this strange error when running the tests. Maybe it's some interference with `pytester` and `pyjs_code_runner`? Maybe @DerThorsten might have more insight

```
<string>:33: DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject filesor modify their metadata. Use the filter argument to control this behavior.
error while evaluating main file: {has_err: true, message: FileNotFoundError: [Errno 44] No such file or dire…ome/web_user/tests'

At:
<string>(2): <module>
, error: FileNotFoundError: [Errno 44] No such file or dire…ome/web_user/tests'
```

IsabelParedes:
So two options here:
1. Remove `pyjs` from run dependencies but keep in host dependencies. This should make the test pass and it makes maintenance a little easier.
2. Or remove the test. If `pyjs` is strictly necessary to be a run dependency (which [doesn't seem so](https://github.com/jupyter-server/jupyverse/blob/fc587ab9fd47cc58dec4e1880253d638dffd8985/plugins/kernel_web_worker/fps_kernel_web_worker/kernel_web_worker.py#L3)), then we can remove the test for now and later restore it once the `pyjs` + `pytester` issue is resolved (which is unrelated to this PR).

DerThorsten:
I think this might be related to https://github.com/emscripten-forge/empack/pull/126/files
All the tests files are "mounts" and the PR above changed some behaviour.

But I need to have a more detailed look

DerThorsten:
> So two options here:
> 
> 1. Remove `pyjs` from run dependencies but keep in host dependencies. This should make the test pass and it makes maintenance a little easier.
> 2. Or remove the test. If `pyjs` is strictly necessary to be a run dependency (which [doesn't seem so](https://github.com/jupyter-server/jupyverse/blob/fc587ab9fd47cc58dec4e1880253d638dffd8985/plugins/kernel_web_worker/fps_kernel_web_worker/kernel_web_worker.py#L3)), then we can remove the test for now and later restore it once the `pyjs` + `pytester` issue is resolved (which is unrelated to this PR).

the `pytester` absolutely needs pyjs as a run dependency

DerThorsten:
removing the wrong  `pyjs<3` in the recipe fixed it

davidbrochart:
> 2\. If `pyjs` is strictly necessary to be a run dependency (which [doesn't seem so](https://github.com/jupyter-server/jupyverse/blob/fc587ab9fd47cc58dec4e1880253d638dffd8985/plugins/kernel_web_worker/fps_kernel_web_worker/kernel_web_worker.py#L3))

Why doesn't it seem so?

DerThorsten:
@davidbrochart jeah one could add a `pyjs<4` but I dont have a strong opinion here

davidbrochart:
@IsabelParedes I have to ignore types as `pyjs` is not an installed dependency for `mypy` in the non-WASM world.

davidbrochart:
BTW @DerThorsten you kept `pyjs >=2,<3` in the host requirements?

IsabelParedes (approved):
Looking good!

PR #2614, opened 2025-07-28
Fix some missing licenses information

arjxn-py:
Adds licence and license_file info (4 at a time)


arjxn-py:
cc: @martinRenou, Do i also need to bump/change build number?

PR #2628, opened 2025-07-31
Add license info for more recipes

arjxn-py:


IsabelParedes:
A lot of R packages use the licenses included in the `r-base` package, you can find them here https://github.com/wch/r-source/tree/trunk/share/licenses . This is why they are not included in the tar ball of those R packages. For those cases, it's a good idea to include the license file directly in the recipe directory.

PR #2645, opened 2025-08-06
Switch `run_modularized` to be `noarch`

ianthomas23:
Switch `run_modularized` to be `noarch` instead of platform-specific (currently it is `linux-64`). The packaged code is pure JavaScript so it makes sense for it to be `noarch`.

Thanks to @DerThorsten for the specific changes.


ianthomas23:
Looks good to me:
```
Archive written to '/home/runner/work/recipes/recipes/output/noarch/run_modularized-0.1.2-h267e887_3.tar.bz2'
```

PR #2733, opened 2025-09-01
Second implementation of dynare_preprocessor python library.

albop:
Here is the second version of the preprocessor (thanks to @ousema-bouaneni)

New:
- the source now contains a basic dynare object, with the ability to compute derivatives at any order
- the build usses the meson file from the source, instead of an adhoc one.


albop:
@DerThorsten : can you have a look please ? Locally it passes the tests already

DerThorsten:
looks good to me! Thanks!

PR #2974, opened 2025-09-18
Added recipe for fast-histogram

astrofrog:
I would like to add a recipe for fast-histogram as it is used by a number of other packages and it contains C extensions.


IsabelParedes (approved):
Thank you!

PR #3190, opened 2025-10-17
update protobuf to 6.33.0

trungleduc:


trungleduc:
@DerThorsten I don't know what i'm doing here, but the CI is green 🤞

PR #3387, opened 2025-10-22
Add sparrow_ipc recipe

Alex-PLACET:


anutosh491 (approved):
LGTM

Alex-PLACET:
@DerThorsten Can you take a look on this PR ?

recipes/recipes_emscripten/sparrow-ipc/build.sh
```diff
@@ -0,0 +1,31 @@
...
+emcmake cmake .. \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_FIND_ROOT_PATH=$PREFIX \
+  -DCMAKE_INSTALL_PREFIX=$PREFIX \
+  -Dflatbuffers_DIR=$PREFIX/lib/cmake/flatbuffers
+
+# Build step
+emmake make -j8
+
+emcc libsparrow-ipc.a \
```
DerThorsten: why not compiling it as shared library in the first place?

recipes/recipes_emscripten/sparrow-ipc/recipe.yaml
```diff
@@ -0,0 +1,50 @@
...
+  build:
+  - ${{ compiler("cxx") }}
+  - cmake
+  - flatbuffers >=25.9.0,<25.10.0
+  host:
+  - flatbuffers >=25.9.0,<25.10.0
+  - sparrow
+  run:
+  - flatbuffers >=25.9.0,<25.10.0
+  - sparrow
```
DerThorsten: is this needed to add sparrow? If so, that indicates a missing run export of sparrow

DerThorsten: same for flatbuffers

recipes/recipes_emscripten/sparrow-ipc/build.sh
DerThorsten: I think you dont need to change the cmakelist, you create that file:
```cmake
# overwriteProp.cmake
set_property(GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS TRUE) # does not need to be global :)
```

and in the build script you add:
```
# CLI
... -DCMAKE_PROJECT_INCLUDE=overwriteProp.cmake
```

This should make it possible to have a shared library (at least when the library type is not explicity marked as static)

anutosh491: Ahh yes we can do this !

Alex-PLACET: I get this cmake error:

> ADD_LIBRARY called with SHARED option but the target platform does not support dynamic linking.  Building a STATIC library instead.

IsabelParedes:
@DerThorsten is this OK to merge?

PR #3943, opened 2025-11-28
add xeus-ocaml for 4-x

anutosh491:


anutosh491:
I'm sure you can try all kernels in the same env with latest jupyterlite-xeus and fetching all kernels from the 4-x branch !!!

PR #4134, opened 2025-12-17
kiwisolver: fix bug in dynamic version detection

MMesch:
This is a fix with test for the following issue:

- Dynamic version detection yields version 0.0.0.0 when assessed through pkg_resources


martinRenou (approved):
LGTM!

PR #4141, opened 2025-12-18
`clarabel`: Add `cffi` and `scipy` as run dependencies

jjerphan:
See: https://github.com/conda-forge/clarabel-feedstock/pull/17


martinRenou (approved):
Thanks!

PR #4149, opened 2025-12-19
Update sparrow recipe version to 2.0.0 [4-x]

anutosh491:


anutosh491:
Merging and trying it out with xeus-cpp 4-x

PR #4218, opened 2026-01-10
add click

trungleduc:


martinRenou:
Thanks @trungleduc ! I'm a bit confused by the PR, click is already installable in an emscripten environment since it's a noarch package on conda-forge. I just tried and successfully installed it, why do you need to add it to emscripten-forge?

martinRenou:
I just published an empty `__unix` package, that seems to work nicely, we can now install `click=8`

PR #4228, opened 2026-01-12
add zarr tests

MMesch:
Basic zarr tests as in this PR are currrently failing. I'm opening to iterate on it and get it running. Currently this is based on the latests zarr.

Would be itnerest to hear feedback from @DerThorsten and @davidbrochart .

### DIFF
--- a/recipes/recipes_emscripten/zarr/patches/0001-Add-WASM-compatible-sync-for-Pyodide.patch
+++ b/recipes/recipes_emscripten/zarr/patches/0001-Add-WASM-compatible-sync-for-Pyodide.patch
@@ -1,0 +1,100 @@
+From e1b0d4c48b209936ab5a0b9b6ccdda82067f867a Mon Sep 17 00:00:00 2001
+From: Agriya Khetarpal <74401230+agriyakhetarpal@users.noreply.github.com>
+Date: Sun, 17 May 2026 21:09:09 +0530
+Subject: [PATCH 1/3] Add WASM-compatible sync() for Pyodide
+
+Pyodide uses WebLoop as its asyncio event loop, which registers itself as
+the running loop at process start. zarr's sync() previously spawned a
+background I/O thread with its own event loop, which is incompatible with
+WASM's single-threaded model and causes RuntimeErrors when called from the
+browser.
+
+On WASM, sync() now delegates to the already-running WebLoop via
+run_until_complete(). If JSPI is not available the loop returns the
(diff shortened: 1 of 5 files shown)


davidbrochart:
It looks like in v3 they use a thread pool executor internally, which was not the case in v2.
I don't know if we can still use v2. At least v3 mentions pyodide compatibility: https://github.com/zarr-developers/zarr-python/releases/tag/v3.0.0.

MMesch:
recipe is pinned to <v3 https://github.com/pyodide/pyodide-recipes/blob/main/packages/zarr/meta.yaml

@DerThorsten we should maybe do the same?

agriyakhetarpal:
Hi! You may refer to https://github.com/zarr-developers/zarr-python/pull/1903 and https://github.com/pyodide/pyodide-recipes/pull/584 for Zarr v3 support.

MMesch:
nice, will have a look.

MMesch:
Reset this branch to a fresh cut at zarr v3.

- zarr 3.1.3 → 3.2.1
- Ported the three WASM patches from pyodide-recipes#584 (upstream zarr-python#1903, still open)
- Swapped `crc32c` → `google-crc32c` (zarr 3.2 imports it directly), dropped unused `asciitree`
- Tests: import still passes, three sync-API roundtrips are `xfail`

The reason for the xfails, and a question for you @DerThorsten: zarr 3's sync API seems to need JSPI. Without it every sync call blows up with:

```
RuntimeError:
Cannot use synchronous Zarr API in browser-based environments without JSPI.
Zarr requires JavaScript Promise Integration (JSPI) to block for asynchronous
operations, which is not supported or enabled in your current environment.
...
```

Pyodide flipped JSPI on by default in 0.28. Pyjs's CMakeLists doesn't seem to have `-sJSPI` (or `-sASYNCIFY`). Could we turn it on? Would that raise the browser floor much? Happy to open an issue on pyjs and prototype if useful.

Once JSPI is around the xfails should flip to XPASS on their own.

agriyakhetarpal:
> Could we turn it on? Would that raise the browser floor much?

For the record, JSPI is supported in Node.js and Chrome, and now Firefox 153, released in July, also supports it: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/153. Safari still doesn't have it, but I suppose you could make the decision based on your browser usage numbers.

davidbrochart:
I think only supporting the async Zarr API (which doesn't require JSPI support IIUC) would already be nice.
What do you think @MMesch ?

PR #4280, opened 2026-01-20
Add leafmap and whiteboxgui recipes (#4279)

martinRenou:


martinRenou:
Ah yeah, whitebox cannot work as-is and indeed would need some amount of work to make it work I guess. 

But the initial intent of that PR is to get `import leafmap` to work without error. I understand `leafmap` falls back to an ipywidgets-base implementation of whitebox if whiteboxui cannot be used. That all that matters to me right now honestly. 

So if we're fine with not having an `import whiteboxui` test in the recipe, and maybe just test the files are installed, I'd be fine with getting the PR in. 

What do you think?

recipes/recipes_emscripten/whiteboxgui/recipe.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - pip
+  host:
+    - python
+  run:
+    - python
+    - ipyfilechooser
+    - ipytree
+    - ipywidgets
+    - whitebox
+    - setuptools
```
martinRenou:
```suggestion
    - setuptools<82
```

PR #4330, opened 2026-01-28
Update clang-resource-headers version to 21.1.8

anutosh491:


anutosh491:
Ahhh okay got it. Sorry for ignoring this information. Shall be making a PR soon and shall tag you for review. Thanks for pointing this out 🙏

recipes/recipes_emscripten/clang-resource-headers/recipe.yaml
IsabelParedes: @anutosh491 If this is only taking some headers from the LLVM package, it would be better to make the LLVM recipe a multi-output recipe. That way they would both be updated together and the LICENSES would not need to be included separately.

PR #2668, opened 2025-08-13
[WIP] Update rustworkx to 0.17.1

IvanIsCoding:
Follow up of #2327

This switches to a publicly released version from the main repository that supports building with emscripten

PR #4350, opened 2026-01-29
Add rustworkx

IsabelParedes:
Supersedes:
- #3622
- #2668


IvanIsCoding:
I think the problem was that #2668 didn't use `rust-nightly`. That is only needed for WASM due to `-Z build-std`, so it is easy to miss. Thanks for fixing it!

PR #4420, opened 2026-02-04
Try polars

AntoinePrv:


recipes/recipes_emscripten/polars/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+      script:
+        file: build_polars_runtime
+      python:
+        version_independent: true
+    requirements:
+      build:
+        - python
+        - cross-python_${{ target_platform }}
+        - cffi
+        - rust
```
IsabelParedes: I recommend using `rust-nightly`. We need it for all Rust packages to enable `wasm-exceptions`

PR #4524, opened 2026-02-11
Reduce astropy package size

martinRenou:
This reduces the package content (once unzipped) by 28.564007MB


martinRenou:
CI passed! I'm merging

PR #4544, opened 2026-02-11
Reduce future package size

martinRenou:
This reduces the package content (once unzipped) by 1.943886MB


martinRenou:
We're testing `import future.tests` which is explicitly filtered out. Either we stop testing that import (if it's not really relevant) or we simply close this PR.

PR #4553, opened 2026-02-11
Reduce markupsafe package size

martinRenou:
This reduces the package content (once unzipped) by 0.028173MB


martinRenou:
CI passed! I'm merging

PR #4686, opened 2026-02-12
Split astropy into astropy-base and astropy

astrofrog:
On conda-forge, astropy now follows the same pattern as matplotlib and defines astropy-base (the real package) and astropy (a meta-package that also includes optional dependencies).

This is now causing problems if I try and install noarch packages from conda-forge that depend on astropy-base because they cannot properly resolve the dependency:

```
[13:44:29]: Cannot solve the request because of: glue-jupyter * cannot be installed because there are no viable options:
├─ glue-jupyter 0.26.0 would require
│  └─ glue-core >=1.25.0, which cannot be installed because there are no viable options:
│     └─ glue-core 1.25.0 would require
│        └─ astropy-base >=4.0, which cannot be installed because there are no viable options:
│           └─ astropy-base 4.0 | 4.0.1.post1 | 4.0.2 | 4.0.3 | 4.0.4 | 4.0.6 | 4.1 | 4.2 | 4.2.1 | 4.3 | 4.3.post1 | 4.3.1 | 5.0 | 5.0.1 | 5.0.2 | 5.0.3 | 5.0.4 | 5.0.5 | 5.0.6 | 5.1 | 5.1.1 | 5.2 | 5.2.1 | 5.2.2 | 5.3 | 5.3.1 | 5.3.2 | 5.3.3 | 5.3.4 | 6.0.0 | 6.0.1 | 6.1.0 | 6.1.1 | 6.1.2 | 6.1.3 | 6.1.4 | 6.1.5 | 6.1.6 | 6.1.7 would require
│              └─ astropy ==4.0, for which no candidates were found.
```

What it's finding are 'fake' astropy-base packages up to 6.1.7 which pre-date the split and just depend on the astropy pacakge with the same version.

I think the solution is to also split astropy here into astropy-base and astropy, as done in this PR.

What I'm not sure about is what to do about the astropy metapackage. We could simply change the recipe to *just* build astropy-base and then let the astropy metapackage come from conda-forge?

If we keep the metapackage here though, we need to figure out what to do about dependencies - should we copy the same dependencies as on conda-forge, or will that be too many?

cc @martinRenou since I see you touched the astropy recipe recently

### DIFF
--- a/emci/rattler_build.py
+++ b/emci/rattler_build.py
@@ -15,8 +15,12 @@ def build_with_rattler(recipe=None, recipes_dir=None, emscripten_wasm32=False, s
         raise ValueError("recipe or recipes_dir must be set")
     elif recipe is not None:
         cmd.extend(["--recipe", str(recipe)])
+        if recipe == "thrift":
+            cmd.extend(["--experimental"])
     elif recipes_dir is not None:
         cmd.extend(["--recipe-dir", str(recipes_dir)])
+        if "thrift" in os.listdir(recipes_dir):
+            cmd.extend(["--experimental"])
 
     cmd.extend(["--skip-existing", skip_existing])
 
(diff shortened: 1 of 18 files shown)


martinRenou:
Sure thing! Can you look into opening a PR in the `emscripten-3x` branch?

PR #4724, opened 2026-02-12
[Emscripten 3.x] Reduce scikit-image package size

martinRenou:
This reduces the package content (once unzipped) by 2.270571MB


martinRenou:
Closing as the gain is not worth having a look

PR #4758, opened 2026-02-13
[Emscripten 3.x] Reduce boost-cpp package size

martinRenou:
This reduces the package content (once unzipped) by 1.160173MB


martinRenou:
Just a linter warning, let's get this in

PR #4835, opened 2026-02-14
.github/workflows/build_recipes.yaml: Upload packages as an artifact

mkoeppe:
So that contributors can download and inspect or test the built packages on PRs.

Preview of the workflow run: https://github.com/emscripten-forge/recipes/actions/runs/22009704864/job/63600884353?pr=4446#step:8:39


martinRenou (approved):
Thanks! Agreed, we saw with @arjxn-py the other day that this was missing and could be handy

PR #4860, opened 2026-02-17
Astropy, matplotlib, clang-resource-headers, fps as noarch

martinRenou:

---

## PR Formatting
- [ ] PR description includes:
  - Version being added/updated

## Package Details
- **Package Name**: 
- **Version**:


anutosh491:
> Let's fix the clang-resources license as part of this PR

Sorry Martin, I might have ignored it in the past. But clang-resource-headers is nothing but a set of header files that xeus-cpp needs and this comes from LLVM/Clang. It's just that xeus-cpp doesn't depend on clang directly and depends on cppinterop.

So what we do is -> copy these headers from llvm -> add as a runtime dep in cppinterop -> use in xeus-cpp.

So it's like a dummy package which probably should share the same licencse as LLVM.

IsabelParedes (approved):
Looks good!

PR #4949, opened 2026-02-24
Avoid propagation of R dependencies in non R packages for Arrow

Alex-PLACET:
Avoid the propagation of R dependencies in non R packages for Arrow packages like Python or C++


recipes/recipes_emscripten/arrow/recipe.yaml
IsabelParedes:
```suggestion
      - r-base-abi
      - r-assertthat
```

PR #4970, opened 2026-02-26
Rename msgpack to msgpack-python

astrofrog:
This is neither just updating a version or adding a new package, it is renaming msgpack to msgpack-python to match conda-forge. To preserve compatibility with packages using ``msgpack``, I have made it so there is an output from the recipe which is a msgpack metapackage that pins to msgpack-python.

This should be checked carefully to make sure it will work as intended!

## PR Formatting
- [ ] PR description includes:
  - Version being added/updated

## Package Details
- **Package Name**: msgpack/msgpack-python
- **Version**: 1.1.2


IsabelParedes (approved):
Great. Thanks!

PR #5507, opened 2026-05-20
Add flatcc

Alex-PLACET:
## Package Details
- **Package Name**: flatcc
- **Version**: 896db54787e8b730a6be482c69324751f3f5f117

### DIFF
--- a/recipes/recipes_emscripten/flatcc/recipe.yaml
+++ b/recipes/recipes_emscripten/flatcc/recipe.yaml
@@ -1,0 +1,54 @@
+context:
+  name: flatcc
+  version: "896db54787e8b730a6be482c69324751f3f5f117"
+
+package:
+  name: ${{ name }}
+  version: ${{ version }}
+
+source:
+  url: https://github.com/dvidelabs/flatcc/archive/${{ version }}.tar.gz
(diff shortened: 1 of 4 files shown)


recipes/recipes_emscripten/flatcc/recipe.yaml
```diff
@@ -0,0 +1,54 @@
+context:
+  name: flatcc
+  version: "896db54787e8b730a6be482c69324751f3f5f117"
```
IsabelParedes, line 3: Why this version and not an official release? v0.6.1?

The bot will not be able to update this version.

IsabelParedes (changes requested):

recipes/recipes_emscripten/flatcc/recipe.yaml
Alex-PLACET: There is no release for a while but the contribution to the main branch are still active. And the recipe which will use this recipe currently use a reference committed after the last release.

IsabelParedes: Fair. It seems they're overdue for a new release.

https://github.com/dvidelabs/flatcc/issues/368

IsabelParedes (approved):
Thanks!

mikkelfj:
FYI, flatcc has a new release.

PR #5620, opened 2026-06-08
Add proxsuite-cpp

Alex-PLACET:
## Package Details
- **Package Name**: proxsuite-cpp
- **Version**: 0.7.3

### DIFF
--- a/recipes/recipes_emscripten/proxsuite/recipe.yaml
+++ b/recipes/recipes_emscripten/proxsuite/recipe.yaml
@@ -1,0 +1,72 @@
+context:
+  name: proxsuite-cpp
+  version: 0.7.3
+
+package:
+  name: ${{ name }}
+  version: ${{ version }}
+
+source:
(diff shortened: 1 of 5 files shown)


recipes/recipes_emscripten/proxsuite/recipe.yaml
```diff
@@ -0,0 +1,72 @@
+context:
+  name: proxsuite-cpp
```
JohanMabille, line 2: The name is just proxsuite.

Alex-PLACET: On conda, the python package use the name "proxsuite", as we try to follow the same names, I postfix the cpp package with "cpp"

JohanMabille: Then you cannot use the name directly to get the archive, because it gives an invalid URL.

EDIT: Ha I just saw you changed that.

JohanMabille (approved):

PR #5673, opened 2026-06-12
protobuf: Add cmake config file ans fix linking

Alex-PLACET:
## Package Details
- **Package Name**: protobuf
- **Version**: 7.35.0

Update the recipe to add cmake config file and fix the linking

### DIFF
--- a/recipes/recipes_emscripten/libprotobuf/build.sh
+++ b/recipes/recipes_emscripten/libprotobuf/build.sh
@@ -39,5 +39,14 @@ emcmake cmake -GNinja \
 
 ninja install
 
+# Protobuf's static runtime also depends on utf8_range support archives.
+find . -name "libutf8_range.a" -exec cp {} "${PREFIX}/lib/" \; 2>/dev/null || true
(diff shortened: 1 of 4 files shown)


recipes/recipes_emscripten/libprotobuf/build.sh
```diff
@@ -39,5 +39,14 @@ emcmake cmake -GNinja \
 
 ninja install
 
+# Protobuf's static runtime also depends on utf8_range support archives.
+find . -name "libutf8_range.a" -exec cp {} "${PREFIX}/lib/" \; 2>/dev/null || true
+find . -name "libutf8_validity.a" -exec cp {} "${PREFIX}/lib/" \; 2>/dev/null || true
```
JohanMabille: I don't think we need this since the existence of these files is already tested in the test section of the recipe.yaml file.

Alex-PLACET: You are right, fixed

JohanMabille (approved):

PR #5679, opened 2026-06-15
Fix sentencepiece by not vendoring protobuf

Alex-PLACET:
## Package Details
- **Package Name**: sentencepiece
- **Version**: 0.2.1

Fix usage of sentencepiece by not vendoring protobuf


IsabelParedes (approved):
Thank you!

PR #5708, opened 2026-06-19
Fix cairo font scaling in r-base

adamblake:
`cairo_set_font_size()` is a no-op on `emscripten-wasm32`, so all plot text renders at one size and `cex`/`pointsize` changes (e.g. larger titles) are ignored. Pin it to 1.0 and apply the requested size as a CTM scale via a shared `FT_getFontScale()` helper: the measure functions multiply returned metrics by the scale and `Cairo_Text` folds it into its existing transform stack.

Verified end-to-end on r-base 4.5.3; identical patch/source on 4.6.0.

## Checklist for **updating** a package

---

## PR Formatting
  - Version being added/updated

## Package Details
- **Package Name**: r-base
- **Version**: 4.6.0 (build 1)

## Build Notes
Adds one patch to r-base's `grDevices` cairo backend to fix font scaling on emscripten. No new dependencies; version unchanged, build number bumped `0 → 1`. The affected source (`grDevices/src/cairo/cairoFns.c`) is identical on 4.5.3 and 4.6.0, so the patch is version-portable.

### DIFF
--- a/recipes/recipes_emscripten/r-base/patches/0013-Fix-cairo-font-scaling-in-emscripten.patch
+++ b/recipes/recipes_emscripten/r-base/patches/0013-Fix-cairo-font-scaling-in-emscripten.patch
@@ -1,0 +1,115 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Adam Blake <adam@ablake.dev>
+Date: Thu, 19 Jun 2026 00:00:00 +0000
+Subject: [PATCH] Fix cairo font scaling under emscripten
+
+Under emscripten-wasm32, cairo_set_font_size() has no effect: the cairo
+font matrix stays at a fixed scale regardless of the requested size, so
+every string renders at the same size and cex / pointsize changes (for
+example a larger plot title) are silently ignored.
+
+Pin cairo_set_font_size() to 1.0 in FT_getFont and apply the requested
+size as a CTM scale instead. A shared FT_getFontScale() helper selects
+the font, reads back the (stuck) font matrix, and returns size / fm.xx.
+Cairo_MetricInfo and Cairo_StrWidth multiply the returned text metrics by
+that scale so measurement matches rendering; Cairo_Text folds the scale
+into its existing transform stack (translate to the origin, rotate, then
+scale) and draws at (0,0).
+---
+diff --git a/src/library/grDevices/src/cairo/cairoFns.c b/src/library/grDevices/src/cairo/cairoFns.c
+index 50ff723..906f26e 100644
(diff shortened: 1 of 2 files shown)


adamblake:
@IsabelParedes anything I can do to make this review easier? This would present a major improvement to all R users as currently font-sizes are very tiny (illegible) on all plots (https://github.com/jupyter-xeus/xeus-r/issues/171)

IsabelParedes (approved):
Thank you @adamblake ! This seems to be working pretty well.

agoose77:
@adamblake can I learn more about how you arrived at this solution? Did you look at the underlying reason for `cairo_set_font_size()` behaviour and the way that fontconfig is not configured? I was looking at this recently and I'm curious to see how you approached the problem.

adamblake:
@agoose77 my first thought was to just figure out what was different from desktop R in modern setups, and (on Linux at least) it appears to use `pango` for the layout and sizing. At the time, emscripten-forge didn't have `glib`, `pango`, or `harfbuzz` so I kind of deferred it. 

Eventually I did get the `glib` + `pango` stack compiled, but it was still trapping at run time when you try to create a text plot. It was a `call_indirect` signature mismatch, and I ran into issues addressing that in the single-threaded, dynamic linking environment we have here. It looked like it would need a bunch of `glib` patches, and I felt like things were getting a little too messy so I shifted gears to focus purely on the `cairo` side of things.

My next guess was similar to yours: no `fontconfig` so no fonts to find. I tried bundling DejaVu and a `fonts.conf` and the env vars but it didn't fix anything scaling-wise. (I did end up making the `fontconfig` fix in a separate PR #5867 and when I did that I checked again and it alone does not fix the scaling, it's orthogonal to this patch)

Then, I started looking at the `CAIRO_HAS_FT_FONT` path, enabled FreeType... and that didn't work either. 

So then I just got fed up and started trying to patch things rather than force code paths. At the time, the fonts were all just so tiny that I tried forcing the font size larger in `cairo_set_font_size()` but that didn't do anything either, so the patch is just on the actual scaling, translation, rotation parts.

This patch is only half of it: to fix sizing it bends cairo's transform, but R measures text through that same transform to lay out math — so on emscripten the measurements come back scaled too. Plain text is fine, but composite plotmath (super/subscripts, limits on sums and integrals) positions pieces relative to each other, so the skewed measurements compound into wrong placement. #5867 measures in a clean, untransformed space to fix it.

---

I still don't know the root cause... I do know that webR ships cairo 1.14.12 and text sizes are correct there so maybe there was a regression (we are on 1.18 here). But building that would probably need its own patches so I'm not sure if it would be worth it or not

PR #5726, opened 2026-06-29
Add pytensor

MMesch:
## Package Details
- **Package Name**: pytensor
- **Version**: 3.0.7

## Build Notes
Builds from sdist with `PYODIDE=1` to skip the cython extension (pure-python mode, as described in https://www.pymc-labs.com/blog-posts/pymc-in-browser).

Three patches adapt pytensor for emscripten:
- `0001`: skip loading C source files in `CLinkerOp.__init__` (the `.c` files are stripped from the browser virtual filesystem by the pytester empack config)
- `0002`: use the pure-python `vm` linker instead of `numba` (numba is unavailable on emscripten)
- `0003`: avoid importing `setuptools` in `link/c/exceptions.py` (setuptools is also stripped from the browser filesystem)

I am unsure whether all of these patches are needed or whether there is a more elegant solution for some of them. I have only tested via the test suite, not yet in a real browser. This is a requirement for running pymc in the browser.

A functional test builds and evaluates `x * 2` via `pytensor.function`.

### DIFF
--- a/recipes/recipes_emscripten/pytensor/recipe.yaml
+++ b/recipes/recipes_emscripten/pytensor/recipe.yaml
@@ -70,4 +70,4 @@ about:
 
 extra:
   recipe-maintainers:
-  - IsabelParedes
+  - mmesch
--- a/recipes/recipes_emscripten/pytensor/recipe.yaml
+++ b/recipes/recipes_emscripten/pytensor/recipe.yaml
@@ -1,0 +1,73 @@
+context:
+  name: pytensor
+  version: 3.0.7
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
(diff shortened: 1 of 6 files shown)


recipes/recipes_emscripten/pytensor/recipe.yaml
```diff
@@ -0,0 +1,73 @@
...
+about:
+  homepage: https://github.com/pymc-devs/pytensor
+  license: BSD-3-Clause
+  license_file: LICENSE.txt
+  summary: Optimizing compiler for evaluating mathematical expressions on CPUs and GPUs
+  documentation: https://pytensor.readthedocs.io/en/latest/
+
+extra:
+  recipe-maintainers:
+  - IsabelParedes
```
MMesch:
```suggestion
  - mmesch
```

DerThorsten (approved):

DerThorsten:
thanks!

PR #5754, opened 2026-07-02
add primecount, primesieve

wangyenshu:
### Pre-submission Checks
- [x] Package requires building for `emscripten-wasm32` platform (not a noarch package), in other words, the package requires compilation.

### Recipe Structure
Added `recipes/recipes_emscripten/[package-name]/recipe.yaml` with proper structure:
- [x] `context` section with `version` (and optionally `name`)
- [x] `source` section with:
  - Source URL is valid and points to archive file (`.tar.gz`, `.tar.bz2`, `.tar.xz`, `.tgz`, or `.zip`)
  - Source URL contains `${{ version }}` template for version updates
  - SHA256 hash is correct (verified with `curl -sL <url> | sha256sum`)
  - Patches (if any) are included in `[package-name]/patches/` directory
- [x] `build` section with appropriate script/method
  - Python packages: `${PYTHON} -m pip install . ${PIP_ARGS}`
  - R packages: `$R CMD INSTALL $R_ARGS .`
  - C++ packages: Uses `emcmake`/`emmake` or `emconfigure`/`emmake`
  - Rust packages: Uses `rust-nightly` and `maturin` or appropriate Rust build tool
  - Build number is 0
  - If the script is longer than 3 lines, a *build.sh* is included
- [x] `requirements` section (build, host, run as needed)
- [x] `tests` section
  - Python packages: `test_import_[package].py` file created and referenced
  - C++ packages: Test executable or package_contents test
  - R packages: Package contents test

---

## PR Formatting
  - Version being added/updated

## Package Details
- **Package Name**: primecount, primesieve 
- **Version**: 8.5, 12.14


IsabelParedes:
- Fixes #2879 #2880

recipes/recipes_emscripten/primecount/recipe.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - automake
+  - libtool
+  - make
+  - ${{ compiler('cxx') }}
+
+tests:
+- package_contents:
+    files:
+    - lib/libprimecount.a
+    - lib/libprimesieve.a
```
IsabelParedes: The `primecount` package should not include the `primesieve` library. `primesieve` should instead be a host dependency for `primecount`.

recipes/recipes_emscripten/primesieve/recipe.yaml
```diff
@@ -0,0 +1,40 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+  - autoconf
+  - automake
+  - libtool
+  - make
+  - ${{ compiler('cxx') }}
```
IsabelParedes, line 22: Please add a run_export, same as how it's done on conda-forge

https://github.com/conda-forge/primesieve-feedstock/blob/dfd416839acbd7ce277e5678a552f444c987471a/recipe/recipe.yaml#L22-L23

IsabelParedes (changes requested):
Thank you for your contribution! Only a couple of changes needed.

recipes/recipes_emscripten/primecount/recipe.yaml
wangyenshu: Yes. But primecount bundles the source of primesieve at https://github.com/kimwalisch/primecount/tree/master/lib/primesieve. So libprimesieve.a is there after compiling.

IsabelParedes: There's an option to disable it

https://github.com/kimwalisch/primecount/blob/b414ca6de3d936b4936791ddfc6fa5c15ec07b33/CMakeLists.txt#L10

```cmake
option(BUILD_LIBPRIMESIEVE "Build libprimesieve"                   ON)

...

# By default the libprimesieve dependency is built from source
# (BUILD_LIBPRIMESIEVE=ON). However when packaging primecount
# for e.g. a Linux distro this is not a good idea. For this use
# case it is better to install the libprimesieve package
# (e.g. libprimesieve-dev) and link against that version.

else()
    find_package(primesieve REQUIRED)
```

wangyenshu: @IsabelParedes Done.

IsabelParedes (approved):
Thank you!

PR #5772, opened 2026-07-02
OpenCV python package

Alex-PLACET:
## Package Details
- **Package Name**: OpenCV
- **Version**: 5

Add OpenCV python package


recipes/recipes_emscripten/libopencv/recipe.yaml
```diff
@@ -1,71 +1,129 @@
 context:
   version: 5.0.0
+  name: opencv
```
IsabelParedes, line 3: You can add this variable so that it doesn't need to be repeated

```suggestion
  name: opencv
  build_number: 3
```

```
build:
  number: ${{ build_number }}
```

IsabelParedes (changes requested):
Minor change. Thanks!

IsabelParedes (approved):
Thank you!

PR #5801, opened 2026-07-06
Add rebound recipe (librebound + rebound)

MMesch:
Add emscripten-forge recipes for REBOUND as a single multi-output recipe (following the matplotlib pattern):

**librebound** — C shared library (side module)
- Builds `src/*.c` as a WASM side module with `-sEXPORT_ALL=1`
- Installs to `$PREFIX/lib/librebound.so` and `$PREFIX/include/`
- C test compiles test + source files statically (side module .so cannot be linked by wasm-ld)

**rebound** — Python ctypes wrapper (depends on librebound)
- Patches `setup.py` to skip C extension build on emscripten
- Patches `rebound/__init__.py` to load `librebound.so` from `$PREFIX/lib/` via ctypes
- Patches `rebound/horizons.py` to use `try/except ImportError` around `pyodide.http` import instead of `sys.modules` check
- Patches `src/simulation.c` to remove `emscripten_sleep()` calls in the integration loop and pause poll — these require `-sASYNCIFY` which we do not compile with (no display server); without this patch `sim.integrate()` aborts in a browser context
- Tested with pytester: import + 2-body N-body simulation


recipes/recipes_emscripten/rebound/patches/0003-fix-pyodide-import.patch
```diff
@@ -0,0 +1,18 @@
+--- a/rebound/horizons.py
++++ b/rebound/horizons.py
+@@ -11,12 +11,11 @@
+ 
+ HORIZONSBASEURL = "https://ssd.jpl.nasa.gov/api/horizons.api?"
+ 
+-if "pyodide" in sys.modules:
+-    from urllib.parse import urlencode
++try:
+     from pyodide.http import open_url as urlopen
```
MMesch, line 10: @IsabelParedes @DerThorsten this tries to import pyodide_html by default. I think we only have it via pip and I'm unsure we can use it as direct dependency?

DerThorsten: ```from pyodide.http import open_url as urlopen```
I vagualy remember that this may just work via some pyjs mockery pretenting to be pyodide.

IsabelParedes:
@MMesch Since both recipes have the same source and are related, both packages can be in a single recipe `rebound/recipe.yaml` with multiple outputs.  See `matplotlib` for example
- https://github.com/emscripten-forge/recipes/blob/main/recipes/recipes_emscripten/matplotlib/recipe.yaml

MMesch:
Ok great. I'll do that!

recipes/recipes_emscripten/rebound/patches/0003-fix-pyodide-import.patch
MMesch: I think the try patch is safe now anyway. So resolving.

IsabelParedes (approved):

PR #5816, opened 2026-07-07
Add graphviz

Alex-PLACET:
## Package Details
- **Package Name**: graphviz
- **Version**: 15.1.0

### DIFF
--- a/recipes/recipes_emscripten/graphviz/recipe.yaml
+++ b/recipes/recipes_emscripten/graphviz/recipe.yaml
@@ -1,9 +1,10 @@
 context:
   name: graphviz
   version: 15.1.0
+  build_number: 0
 
 build:
-  number: 0
+  number: ${{ build_number }}
 
 outputs:
   - staging:
@@ -18,7 +19,6 @@ outputs:
         - ninja
         - bison
         - flex
-        - python
         - pkg-config
       host:
         - expat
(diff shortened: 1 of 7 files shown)


anutosh491:
Curious if you tried using it with Jupyterlite ! Did it work ?

recipes/recipes_emscripten/graphviz/recipe.yaml
```diff
@@ -0,0 +1,167 @@
...
+      script: build.sh
+    requirements:
+      build:
+        - ${{ compiler('c') }}
+        - ${{ compiler('cxx') }}
+        - cmake
+        - ninja
+        - bison
+        - flex
+        - python
```
IsabelParedes: Can python be removed?

The build script includes
```bash
-DENABLE_PYTHON=OFF \
```

Python is also already a dependency for the emscripten package, so it shouldn't need to be part of the build dependencies.

This python is also modifying the name of the package, and it shouldn't.

graphviz-15.1.0-<ins>**py313**</ins>h0c76a5c_0

recipes/recipes_emscripten/graphviz/recipe.yaml
```diff
@@ -0,0 +1,167 @@
+context:
+  name: graphviz
+  version: 15.1.0
```
IsabelParedes, line 3:
```suggestion
  version: 15.1.0
  build_number: 0
```

Use this variable so that it doesn't need to be repeated.

IsabelParedes (changes requested):
Minor changes. Thank you!

IsabelParedes (approved):
Thank you!

PR #5847, opened 2026-07-09
Add xerces-c recipe for emscripten-wasm32

tritao:
## Summary
- add a new `xerces-c` recipe for `emscripten-wasm32`
- build a static library with ICU transcoding and disable docs, samples, and tests in the cross build
- add a small upstream patch so an externally provided `CMAKE_CXX_STANDARD` can override Xerces' default `C++14` setting

## Validation
- built locally with `pixi run build-emscripten-wasm32-pkg recipes/recipes_emscripten/xerces-c`
- package tests passed and produced `lib/libxerces-c.a`, headers, and CMake package files

## Context
This was needed to unblock a downstream FreeCAD wasm configure, which was previously failing at `find_package(XercesC)` because `xerces-c` was missing from the target channel.

### DIFF
--- a/recipes/recipes_emscripten/xerces-c/build.sh
+++ b/recipes/recipes_emscripten/xerces-c/build.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+set -exuo pipefail
+
+mkdir -p build
+cd build
+
+emcmake cmake -GNinja \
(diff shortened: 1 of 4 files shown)


IsabelParedes (approved):
Thank you!

PR #5865, opened 2026-07-13
Add ffmpeg cli

Alex-PLACET:
## Package Details
- **Package Name**: ffmpeg
- **Version**: 8.1.1

Add ffmpeg cli

### DIFF
--- a/recipes/recipes_emscripten/ffmpeg/build.sh
+++ b/recipes/recipes_emscripten/ffmpeg/build.sh
@@ -12,19 +12,20 @@ export PYTHON="${BUILD_PREFIX}/bin/python3"
 # Makefile variable of the same name.
 unset SUBDIR
 
-export CFLAGS="-I${PREFIX}/include ${CFLAGS:-}"
-export LDFLAGS="-L${PREFIX}/lib ${LDFLAGS:-}"
+export CFLAGS="-I${PREFIX}/include -pthread ${CFLAGS:-}"
+export LDFLAGS="-L${PREFIX}/lib -pthread ${LDFLAGS:-}"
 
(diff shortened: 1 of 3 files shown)


recipes/recipes_emscripten/ffmpeg/build.sh
```diff
@@ -12,28 +12,44 @@ export PYTHON="${BUILD_PREFIX}/bin/python3"
...
 # llvm-nm avoids the broken packaged emnm Python launcher.
 ./configure \
   --prefix="${PREFIX}" \
-  --target-os=none --arch=x86_32 --enable-cross-compile \
-  --disable-asm --disable-stripping --disable-programs \
+  --target-os=none --arch=wasm --enable-cross-compile \
+  --disable-stripping \
   --disable-doc --disable-debug --disable-runtime-cpudetect \
-  --disable-autodetect --disable-pthreads --disable-w32threads \
+  --disable-autodetect --enable-pthreads --disable-w32threads \
```
DerThorsten, line 24: are threads absolutely needed?

Alex-PLACET: Yes, I can't compile the CLI without it

DerThorsten: does that mean the pthread symbols are only dragged into the binary or are the libs also have them?

Alex-PLACET: I updated the recipe to have two pacakges now, one for the libs and another one for the CLI. I can create two packages which don't share the same core and so have the libs without pthread and the CLI with it

PR #5867, opened 2026-07-14
Fix plotmath rendering in r-base on the cairo FreeType path

adamblake:
On `emscripten-wasm32`, R's cairo device runs on the FreeType/fontconfig text path (there is no pango in the wasm build), and plotmath renders wrong there in two ways: Greek and math symbols come out as tofu, and composite expressions (subscripts, superscripts, and the limits on sums and integrals) are mis-positioned relative to their base.

The fix has four parts:
- Ship an r-base-owned `fonts.conf` and select it with a `FONTCONFIG_FILE` line appended to `etc/Renviron`. fontconfig's wasm-compiled default config path points at the build prefix, which does not exist at runtime; `Renviron` is read unconditionally at startup, before `FcInit` and regardless of `--vanilla`, so this reliably points fontconfig at a config that exists.
- Add `font-ttf-dejavu` run-deps. DejaVu is one family covering Latin, Greek, and math across all four faces, which the FreeType path needs because it does no per-glyph fallback; `fonts.conf` aliases R's families (Helvetica/Times/Courier plus the sans/serif/mono generics) to DejaVu.
- Patch 0016 (`cairo.R`) routes plotmath symbols through Unicode under emscripten instead of the legacy Adobe Symbol Type-1 font, which the FreeType path cannot render.
- Patch 0013 (`cairoFns.c`) gains a six-line addition: `Cairo_MetricInfo` and `Cairo_StrWidth` scale the drawing context around `cairo_text_extents` so text is measured in the same frame it is drawn in. Without it the existing font-scaling patch measures composite plotmath in a different frame than it draws, mis-positioning it.

Verified by building r-base 4.6.1 for `emscripten-wasm32` and rendering a plotmath test image covering all four font faces of Latin and Greek plus sum, integral, subscript, and superscript composites; all render correctly. An A/B against the unmodified 0013 patch confirms the composite fix is necessary and that the six-line form reproduces it exactly.

## Checklist for **updating** a package

---

## PR Formatting
  - Version being added/updated

## Package Details
- **Package Name**: r-base
- **Version**: 4.6.1 (build 1)

## Build Notes
Version unchanged at 4.6.1; build number bumped `0 → 1`. Adds one new patch (`0016`, routes plotmath symbols through Unicode under emscripten) and a six-line addition to the existing `0013` cairo font-scaling patch (measures composite text in the frame it is drawn in). Ships a `fonts.conf` installed into `lib/R/etc` and selected via `FONTCONFIG_FILE` in `Renviron`, plus `font-ttf-dejavu` / `font-ttf-dejavu-sans-mono` (>=2.37, conda-forge) run-deps to supply the four DejaVu faces at runtime.

### DIFF
--- a/recipes/recipes_emscripten/r-base/build.sh
+++ b/recipes/recipes_emscripten/r-base/build.sh
@@ -91,6 +91,13 @@ pushd _build_wasm
     export LINUX_BUILD_DIR=$(realpath ..)/_build_linux
     export WASM_BUILD_DIR=$(pwd)
 
+    # cairo's probe can't link-run under wasm; force it onto the FreeType path.
+    export r_cv_cairo_works=yes
+    export PKG_CONFIG_PATH="$PREFIX/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
+    export CAIRO_CFLAGS="-I$PREFIX/include/cairo -I$PREFIX/include"
+    export CAIRO_LIBS="-L$PREFIX/lib -lcairo -lfreetype -lfontconfig"
+    export CPPFLAGS="${CPPFLAGS:-} -I$PREFIX/include -I$PREFIX/include/cairo"
+
     # RPY_LIBS are appended to $(R_bin_LDADD) when linking the RPY target
     # (see patch 0015). This statically links libpython and its transitive deps
     # into RPY only; the standard R executable is unaffected.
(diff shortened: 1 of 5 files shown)


adamblake:
This should close https://github.com/jupyter-xeus/xeus-r/issues/167

IsabelParedes (approved):
Thank you @adamblake !

IsabelParedes:
> This should close [jupyter-xeus/xeus-r#167](https://github.com/jupyter-xeus/xeus-r/issues/167)

Since R packages from conda-forge have not migrated to the latest R version 4.6.x, these changes will not be reflected in `xeus-r` yet.

If you would like, you could also add these patches to `r-base-v4.5.3/`.

adamblake:
@IsabelParedes sure thing --- I'll add them there too

PR #5951, opened 2026-07-17
Rebuild netcdf4; Rebuild libnetcdf; fix runtime failure due to missing listxattr

MMesch:
Edited after splitting out new PR #6007 . 

# Summary 

Fixes the runtime failure `xarray.open_dataset(...)` → `netCDF4.Dataset` → `nc_open` → `H5DSis_scale` → `Dynamic linking error: cannot resolve symbol H5DSis_scale` in the browser, and the follow-on `RuntimeError: ring type mismatch occurred for cache entry` that surfaces once `H5DSis_scale` is reachable.

# Build Notes

- **netcdf4**: add `-lhdf5_hl` to LDFLAGS. This may not strictly be necessary since it seems to be activated by default. But it's probably safer to add. The bug we observed was related to this library.
- **libnetcdf**: skip the DAOS-container check under `__EMSCRIPTEN__`: netcdf-c 4.10.1 references `listxattr` from `libdispatch/dinfermodel.c`inside the DAOS-container probe, guarded by `HAVE_SYS_XATTR_H`. Emscripten's `<sys/xattr.h>` declares `listxattr` but the runtime doesn't implement it, so the built `.so` fails to load. `configure.ac` seems to have no DAOS knobs. We therefore patch the source to early-return `NC_ENOTNC` under `__EMSCRIPTEN__`. Attempting to set the variable `ac_cv_header_sys_xattr_h=no` produced garbled compile errors in the surrounding `isdaoscontainer()` function under `-std=gnu23` and was abandoned.
- `variant.yaml` bumped: `hdf5: 1.14.6`, `libnetcdf: 4.10.1`.

### DIFF
--- a/recipes/recipes_emscripten/libnetcdf/patches/skip-daos-emscripten.patch
+++ b/recipes/recipes_emscripten/libnetcdf/patches/skip-daos-emscripten.patch
@@ -1,0 +1,25 @@
+DAOS container detection uses listxattr/getxattr which are declared in
+Emscripten's <sys/xattr.h> but not implemented at runtime — the built .so
+fails to load with "cannot resolve symbol listxattr". DAOS makes no sense
+in a WASM sandbox anyway. Short-circuit isdaoscontainer() under
+__EMSCRIPTEN__ so the xattr code is never reached.
+
+--- a/libdispatch/dinfermodel.c
++++ b/libdispatch/dinfermodel.c
+@@ -1594,6 +1594,10 @@
+ static int
+ isdaoscontainer(const char* path)
+ {
++#ifdef __EMSCRIPTEN__
++    (void)path;
++    return NC_ENOTNC;
++#else
+     int stat = NC_ENOTNC; /* default is that this is not a DAOS container */
+ #ifndef _WIN32
+ #ifdef USE_HDF5
+@@ -1669,6 +1673,7 @@
+ #endif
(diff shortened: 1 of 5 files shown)


MMesch:
@IsabelParedes , waiting for your thoughts on this now. After discussing with @DerThorsten these are some points to consider:

- the feclearexcept(FE_INVALID) issue is solved upstream in hdf5 2.x.
- the deduplication problem should be solved generally. But we can use the patch here as hot fix and move ahead with this recipe more quickly.
- the netcdf updates could be split out in a separate PR

IsabelParedes:
> the netcdf updates could be split out in a separate PR

This would be nice. Thanks!

DerThorsten:
> 2\. `libhdf5_hl.so` records NEEDED `libhdf5.so.310.5.1`

@MMesch  isnt that the culprint / root-problem
I would expect that one never explicitly links against the versioned so, just the unversioned so (ie libhdf5.so)
the unversioned so ist then just a symlink to the versioned so, that way one can swap out the symlink and point to a newer version.

@MMesch did you trace the stuff in the browser to see that `libhdf5_hl.so` records NEEDED `libhdf5.so.310.5.1` or whats a good way to see that

MMesch:
@DerThorsten , I split out hdf5 to PR #6007. I'll write a more detailed description what I did in there.

PR #5965, opened 2026-07-19
Add gap, passagemath-gap, passagemath-groups

mkoeppe:
### Pre-submission Checks
- [x] Package requires building for `emscripten-wasm32` platform (not a noarch package), in other words, the package requires compilation.

### Recipe Structure
Added `recipes/recipes_emscripten/[package-name]/recipe.yaml` with proper structure:
- [x] `context` section with `version` (and optionally `name`)
- [x] `package` section with name and version using Jinja2 templates
- [x] `source` section with:
  - Source URL is valid and points to archive file (`.tar.gz`, `.tar.bz2`, `.tar.xz`, `.tgz`, or `.zip`)
  - Source URL contains `${{ version }}` template for version updates
  - SHA256 hash is correct (verified with `curl -sL <url> | sha256sum`)
  - Patches (if any) are included in `[package-name]/patches/` directory
- [x] `build` section with appropriate script/method
  - Python packages: `${PYTHON} -m pip install . ${PIP_ARGS}`
  - R packages: `$R CMD INSTALL $R_ARGS .`
  - C++ packages: Uses `emcmake`/`emmake` or `emconfigure`/`emmake`
  - Rust packages: Uses `rust-nightly` and `maturin` or appropriate Rust build tool
  - Build number is 0
  - If the script is longer than 3 lines, a *build.sh* is included
- [x] `requirements` section (build, host, run as needed)
- [x] `tests` section
  - Python packages: `test_import_[package].py` file created and referenced
  - C++ packages: Test executable or package_contents test
  - R packages: Package contents test
- [x] `about` section with license, homepage, summary

---

## PR Formatting
- [x] PR title follows format: `Add [package-name]` or `Update [package-name] to [version]`
- [x] PR description includes:
  - Version being added/updated
  - Any special build considerations or patches applied

## Package Details
- **Package Name**: gap, passagemath-gap, passagemath-groups
- **Version**: 4.16.0, 10.8.6, 10.8.6

## Build Notes

PR is on top of and supersedes
- https://github.com/emscripten-forge/recipes/pull/5866 by @wangyenshu 

FYI @ChrisJefferson

PR #5891, opened 2026-07-16
add gap-in-the-browser

wangyenshu:
### Pre-submission Checks
- [x] Package requires building for `emscripten-wasm32` platform (not a noarch package), in other words, the package requires compilation.

### Recipe Structure
Added `recipes/recipes_emscripten/[package-name]/recipe.yaml` with proper structure:
- [x] `context` section with `version` (and optionally `name`)
- [x] `package` section with name and version using Jinja2 templates
- [x] `source` section with:
  - Source URL is valid and points to archive file (`.tar.gz`, `.tar.bz2`, `.tar.xz`, `.tgz`, or `.zip`)
  - Source URL contains `${{ version }}` template for version updates
  - SHA256 hash is correct (verified with `curl -sL <url> | sha256sum`)
  - Patches (if any) are included in `[package-name]/patches/` directory
- [x] `build` section with appropriate script/method
  - Python packages: `${PYTHON} -m pip install . ${PIP_ARGS}`
  - R packages: `$R CMD INSTALL $R_ARGS .`
  - C++ packages: Uses `emcmake`/`emmake` or `emconfigure`/`emmake`
  - Rust packages: Uses `rust-nightly` and `maturin` or appropriate Rust build tool
  - Build number is 0
  - If the script is longer than 3 lines, a *build.sh* is included
- [x] `requirements` section (build, host, run as needed)
- [x] `tests` section
  - Python packages: `test_import_[package].py` file created and referenced
  - C++ packages: Test executable or package_contents test
  - R packages: Package contents test
- [x] `about` section with license, homepage, summary

---

## PR Formatting
- [x] PR title follows format: `Add [package-name]` or `Update [package-name] to [version]`
- [x] PR description includes:
  - Version being added/updated
  - Any special build considerations or patches applied

## Package Details
- **Package Name**: gap-in-the-browser
- **Version**: 4.16.0

## Build Notes

This recipe will build gap-in-the-browser. The output should be the same as https://gap-system.github.io/gap-in-the-browser/.

PR #6016, opened 2026-07-24
gap: Add packages, switch to ASYNCIFY=1, add runtime tests

mkoeppe:
Adding tests as suggested in https://github.com/emscripten-forge/recipes/pull/5965#issuecomment-5067956541

FYI @wangyenshu @ChrisJefferson

### DIFF
--- a/recipes/recipes_emscripten/gap/build.sh
+++ b/recipes/recipes_emscripten/gap/build.sh
@@ -2,6 +2,8 @@
 
 set -eux
 
+unset EXCEPTION_HANDLING_FLAGS
+unset EMCC_CFLAGS
 BASEDIR="$(pwd)"
 
 if ! command -v emmake &> /dev/null; then
@@ -57,7 +59,7 @@ if [[ ! -f GNUmakefile ]] || ! grep '/emcc' GNUmakefile > /dev/null; then
     emconfigure ./configure --prefix="${PREFIX}" ABI=32 \
     --with-gmp=${PREFIX} \
     --with-zlib=${PREFIX} \
-    LDFLAGS="-s JSPI -O2"
+    LDFLAGS="-s ASYNCIFY=1 -O2"
 fi;
 
 # Target the Linux/Unix block to redefine shared lib extensions to static
@@ -100,7 +102,7 @@ fi
 
 # The EXEEXT is usually for windows, but here it lets us set GAP's extension,
 # which lets us produce a html page to run GAP in.
-emmake make -j8 LDFLAGS="-lidbfs.js -s JSPI -sTOTAL_STACK=32mb -sINITIAL_MEMORY=2048mb -O2" EXEEXT=".html"
+emmake make -j8 LDFLAGS="-lidbfs.js -s ASYNCIFY=1 -sTOTAL_STACK=32mb -sINITIAL_MEMORY=2048mb -O2" EXEEXT=".html"
 emmake make install-bin install-gaproot install-sysinfo install-headers install-libgap
 
 cp gap.js gap.wasm "${PREFIX}/bin/"
--- a/recipes/recipes_emscripten/gap/build.sh
+++ b/recipes/recipes_emscripten/gap/build.sh
@@ -103,7 +103,7 @@ fi
 # The EXEEXT is usually for windows, but here it lets us set GAP's extension,
 # which lets us produce a html page to run GAP in.
 emmake make -j8 LDFLAGS="-lidbfs.js -s ASYNCIFY=1 -sTOTAL_STACK=32mb -sINITIAL_MEMORY=2048mb -O2" EXEEXT=".html"
-emmake make install
+emmake make -k install || :
 
 cp gap.js gap.wasm "${PREFIX}/bin/"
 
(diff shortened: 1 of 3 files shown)


wangyenshu:
The documentation for gap wasm build is here https://github.com/gap-system/gap/tree/master/etc/emscripten. For a working website, the steps in https://github.com/gap-system/gap/blob/master/etc/emscripten/assemble-website.sh is needed.

ChrisJefferson:
I don't know how much of a test you are happy to run. If a copy of GAP's test directory is available, `Read("tst/testinstall.g");`would be best, but it would take a while. If GASMAN is broken (as I suspect when ASYNCIFY was removed), these tests won't be enough to discover that as they aren't doing enough work to cause the GC to activate and garbage collection to not work correctly.

mkoeppe:
A specific test that exercises the GC (GASMAN) would be most valuable.

wangyenshu:
ASYNCIFY flag works after adding this on the top of the build script.
```
unset EXCEPTION_HANDLING_FLAGS
unset EMCC_CFLAGS
```
See https://github.com/emscripten-forge/recipes/pull/5891/changes.
For the gap recipe, all gap packages are still missing. You need to add
```
bash etc/emscripten/assemble-website.sh
```
The complete website is in "web-example/*".

mkoeppe:
> need to add
> 
> ```
> bash etc/emscripten/assemble-website.sh
> ```

@wangyenshu  I'm not sure if you're on the right track with this website data stuff. The emscripten-forge packages already have access to the emscripten file system, so a standard installation of the GAP packages into `/lib/gap/` should suffice, if I'm not mistaken.

mkoeppe:
> ASYNCIFY flag works after adding this on the top of the build script.
> 
> ```
> unset EXCEPTION_HANDLING_FLAGS
> unset EMCC_CFLAGS
> ```
> 
> See https://github.com/emscripten-forge/recipes/pull/5891/changes.

OK, I'll cherry-pick from there

mkoeppe:
```
Testing commands:
GAP 4.16dev built on reproducible
gap: hmm, I cannot find 'lib/init.g' maybe use option '-l <gaproot>'?
```

wangyenshu:
> > need to add
> > ```
> > bash etc/emscripten/assemble-website.sh
> > ```
> 
> @wangyenshu I'm not sure if you're on the right track with this website data stuff. The emscripten-forge packages already have access to the emscripten file system, so a standard installation of the GAP packages into `/lib/gap/` should suffice, if I'm not mistaken.

You mean installing both gap wasm version and gap host version? In nodejs environment, gap can access the filesystem. But in browser you need to create a virtual filesystem and mount the files.

mkoeppe:
In node and also in the deployment done by XPython

wangyenshu:
Then that will be fine. But users maybe confused if they want to call gap.js directly in browser without xpython. If the gap packages (the pkg directory) is not needed, you can remove this block:
```
if [[ "$PKG_NAME" == "gap" ]]; then
    echo "Building full packages for $PKG_NAME..."
    emmake make bootstrap-pkg-full
elif [[ "$PKG_NAME" == "gap-core" ]]; then
    echo "Building minimal packages for $PKG_NAME..."
    emmake make bootstrap-pkg-minimal
else
    echo "Error: Unknown package name $PKG_NAME"
    exit 1
fi
```

mkoeppe:
> the gap packages (the pkg directory) is not needed

Of course they are needed. It's just still not being installed, to be done.

wangyenshu:
Maybe changing
```
node $PREFIX/bin/gap.js -l /share/gap -c 'quit;'
```
to 
```
node $PREFIX/bin/gap.js -l $PREFIX/share/gap -c 'quit;'
```
would fix the current path issue.

wangyenshu:
> > the gap packages (the pkg directory) is not needed
> 
> Of course they are needed. It's just still not being installed, to be done.
```
emmake make bootstrap-pkg-full
```
This command install the pkg directory. You may need to add
```
cp -r pkg doc "${PREFIX}/share/gap/"
```
to package them. This command will also package doc so that the help command will work.

mkoeppe:
> users maybe confused if they want to call gap.js directly in browser without xpython.

To my understanding, emscripten-forge packages are not in charge of their own deployment

wangyenshu:
What about using preload:
```
EMSCRIPTEN_DIR="$(dirname "$(readlink -f "$(command -v emcc)")")"
python3 "${EMSCRIPTEN_DIR}/tools/file_packager.py" \
  "${PREFIX}/bin/gap.data" \
  --preload "${SRC_DIR}/pkg@/gap/pkg" \
  --preload "${SRC_DIR}/lib@/gap/lib" \
  --preload "${SRC_DIR}/grp@/gap/grp" \
  --preload "${SRC_DIR}/doc@/gap/doc" \
  --preload "${SRC_DIR}/hpcgap@/gap/hpcgap" \
  --js-output="${PREFIX}/bin/gap.data.js"
```
Then the folder structure would be cleaner and the gap's functionality is complete.

mkoeppe:
To my understanding, such deployment concerns are intended to be taken care of by users using https://github.com/emscripten-forge/empack

wangyenshu:
Maybe the unset command also unset the flag needed for interacting with node host filesystem. I am not sure. What is the problem with passagemath-group? Let's see if that problem can be fixed without testing on nodejs.

wangyenshu:
What about adding ```EMCC_CFLAGS=-O2 -g0 -fPIC -msimd128 -sSUPPORT_LONGJMP=wasm``` right after ```unset EMCC_CFLAGS```? Maybe this could fix the test error?

PR #6048, opened 2026-07-26
passagemath-flint: Add runtime tests

mkoeppe:
Repro of the dynamic linking error https://github.com/passagemath/passagemath/issues/1863#issuecomment-5081762662


mkoeppe:
Strangely it fails to reproduce the issue.

PR #6051, opened 2026-07-26
libgc: Use `-fPIC`

mkoeppe:
Trying to address https://github.com/passagemath/passagemath/issues/1863#issuecomment-5071605378 @wangyenshu


mkoeppe:
Apparently not needed; tested in #6052

PR #6086, opened 2026-07-28
Fix SciPy gees wrapper ABI

akhmerov:

---

## PR Formatting
  - Version being added/updated

## Package Details
- **Package Name**: scipy
- **Version**: 1.18.0

## Build Notes
This fixes the following problem.

The Emscripten build of SciPy crashes in `scipy.linalg.schur` with:

```text
RuntimeError: null function or function signature mismatch
```

Minimal reproduction:

```python
import numpy as np
from scipy.linalg import schur

for a in (
    np.array([[1.0, 2.0], [3.0, 4.0]]),
    np.array([[1.0, 2.0j], [3.0j, 4.0]]),
):
    t, z = schur(a)
    np.testing.assert_allclose(z @ t @ z.conj().T, a)
```

The regression was introduced in commit 3b6da57a. Its F2PY patch added the hidden Fortran character-length arguments `len0,len1` to the real and complex `gees` wrappers, but retained the existing trailing `1,1`. This passed four length arguments through a function pointer expecting two, causing an Emscripten ABI signature mismatch.

The fix removes the redundant `1,1` and corresponding prototype entries, leaving:

```c
(*f2py_func)(..., &info, len0, len1)
```

for the `sgees`, `dgees`, `cgees`, and `zgees` wrappers. A regression test now covers both real and complex `scipy.linalg.schur`.


ianthomas23 (approved):
Thanks @akhmerov

PR #6097, opened 2026-07-29
Add libwgpu-native

Alex-PLACET:
## Package Details
- **Package Name**: libwgpu-native
- **Version**: 29.0.1.1


Alex-PLACET:
I finally figure out that we don't need this recipe as it's not used in py-wgpu for wasm backend

PR #6099, opened 2026-07-30
Add flit core

Alex-PLACET:
## Package Details
- **Package Name**: flit-core
- **Version**: 3.12.0


IsabelParedes (changes requested):
This package is already on conda-forge.

> ### Pre-submission Checks
> - [ ] Package requires building for emscripten-wasm32 platform (not a noarch package), in other words, the package requires compilation.

PR #6201, opened 2026-08-08
Update xeus-ocaml from 0.2.8 to 0.2.10, OCaml 5.5.0, fix opam root

davy39:
- Bump version to 0.2.10 with new sha256
- Use a private OPAMROOT and unset OPAMSWITCH so opam init fetches the real opam repository (the conda-forge opam root only ships an empty repo, which broke 'opam install dune' with 'No package named dune found')
- Build with OCaml 5.5.0 to match the v0.2.10 tag
- Unset LDFLAGS before cmake: native -Wl flags are rejected by wasm-ld

### DIFF
--- a/recipes/recipes_emscripten/xeus-ocaml/build.sh
+++ b/recipes/recipes_emscripten/xeus-ocaml/build.sh
@@ -9,7 +9,7 @@ set -ex
   unset CC CXX CFLAGS CXXFLAGS LDFLAGS OPAMSWITCH
   export OPAMROOT=$SRC_DIR/opam_root
   cd ocaml
-  opam init --disable-sandboxing --no --compiler=5.4.0
+  opam init --disable-sandboxing --no --compiler=5.5.0
   opam install dune
   opam exec -- dune pkg lock
   opam exec -- dune build --profile release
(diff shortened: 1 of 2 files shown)


IsabelParedes (changes requested):
I merged #6200 first so that all versions are made available. Would you mind rebasing this?

Thank you!

davy39:
That's done.
Thanks

IsabelParedes (approved):
Perfect. Thank you!

anutosh491:
xeus-ocaml 5.5.0 looks broken on emscripten-forge !?

Some pics from the demo notebook

image
image
image

@davy39 would you like to try it out for me and confirm you see the same ? Once we know it's broken we can look into what's wrong !

davy39:
Hello, 
Thanks for reporting.

My [build](https://davy39.github.io/xeus-ocaml/lab/index.html
) doesn't have those errors. I'm gonna investigate by installing emscripten build locally.

First errors (List.) are normal, this code is just to test completion and doc.

From the ocamlgraph require error, my first guess would be a problem with xocaml getting the dependencies url.

In my build, there is indeed a new bug with fs streaming (input_line) , look like a pos method missing that I've to add.

I'll try to get into it this week.

PR #6239, opened 2026-08-11
shared libproj; tests and rebuild for pyproj

MMesch:

- ~[] Or reset build number to 0 if updating the package to a newer version~

---

## PR Formatting
- [ ] PR title follows format: `Add [package-name]` or `Update [package-name] to [version]` (unsure since this isn't a package update)
  - Version being added/updated

## Package Details
- **Package Name**: libproj, pyproj
- **Version**: unchanged versions

## Build Notes

- I build proj as shared library in this PR. This seems to reduce size from 118Mb (pyproj) to 1.8MB (pyproj) + 4.6MB (libproj) since pyproj otherwise embeds and duplicates much of proj's functionality across it's ~10 .so files. A shared build therefore seems to be the right call here.
- I deactivated the SONAME in the build as we did in PR [#6001](https://github.com/emscripten-forge/recipes/pull/6007), for the same reason that the wasm linker identifies libraries by full name without understanding of semantic version or deduplication of symlinks. This means that two libraries loading the *same* proj in two different ways, end up with two instances of it which is usually unintended and throws exceptions if each instance handles its own state.
- I added a set of additional tests to this.
- I also remove pip from the host build dependencies since it seems it is already provided by cross_python. The additional pip threw a build error (locally on my machine), likely because of duplication. Removing it seems to work fine.

### DIFF
--- a/recipes/recipes_emscripten/proj/build.sh
+++ b/recipes/recipes_emscripten/proj/build.sh
@@ -9,7 +9,7 @@ cd build
 # to build without curl we need to disable projsync too
 emcmake cmake ${CMAKE_ARGS} .. \
       -DCMAKE_BUILD_TYPE=Release \
-      -DBUILD_SHARED_LIBS=OFF \
+      -DBUILD_SHARED_LIBS=ON \
       -DBUILD_APPS=OFF \
       -DBUILD_TESTING=OFF \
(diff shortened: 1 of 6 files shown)


IsabelParedes (approved):
Thank you!

PR #6264, opened 2026-08-12
Run OpenBLAS test suite

jjerphan:
## Package Details
- **Package Name**: openblas
- **Version**: 0.3.34

## Build Notes
Rebased onto main (OpenBLAS 0.3.34). Reuses the Emscripten/Flang patches from main (0001–0008). New patches 0009–0012 adapt the test suite for wasm: skip fork-based utests, run under Node, and align ctest F77/f2c wrappers with the wasm ABI.

Fortran BLAS drivers (`test/` and `ctest/*.f`) are skipped because flang 22 segfaults on wasm; the build runs utest plus CBLAS levels 1–3 with `NOFORTRAN=1`.

### DIFF
--- a/recipes/recipes_emscripten/openblas/recipe.yaml
+++ b/recipes/recipes_emscripten/openblas/recipe.yaml
@@ -12,6 +12,7 @@ source:
   patches:
   # Branch containing these patches is v0.3.34-emforge at https://github.com/jjerphan/OpenBLAS
   # Use `git format-patch v0.3.34..v0.3.34-emforge` to create patch files.
+  # Patches 0009-0012 adapt the OpenBLAS test suite for emscripten/wasm.
   - patches/0001-Remove-march-flags.patch
   - patches/0002-Skip-linktest.patch
   - patches/0003-Add-emscripten-as-supported-arch.patch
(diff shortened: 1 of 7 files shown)


ianthomas23 (approved):

recipes/recipes_emscripten/openblas/recipe.yaml
```diff
@@ -20,6 +21,10 @@ source:
   - patches/0006-Int-to-void-return-types-of-lapack-functions-used-by.patch
   - patches/0007-Enable-LAPACK_FORTRAN_STDLEN_END-for-lapacke-calling.patch
   - patches/0008-Clear-FEXTRALIB-when-linking-shared-library.patch
+  - patches/0009-Skip-fork-based-utests-for-emscripten.patch
+  - patches/0010-Run-OpenBLAS-tests-under-node-for-emscripten.patch
+  - patches/0011-Return-int-from-ctest-F77-wrappers-for-wasm-ABI.patch
+  - patches/0012-Use-int-return-for-f2c-Subroutine-decls-in-ctest.patch
 
 build:
   number: 1
```
IsabelParedes, line 30: @jjerphan 

Looks like you forgot to bump the build number. This package was not uploaded.

```
Uploading openblas-0.3.34-h69b678e_1.tar.bz2 for emscripten-wasm32
Error:   × package already exists (HTTP 409)
```

jjerphan: Ah, this is what I intended (AFAIK, we don't need another build published, just the test suite to be run at build time; is this right?).

I can open another PR to bump the build number if needed.

IsabelParedes: It's fine for now. But generally it's better to always bump the build number and publish the package even if it doesn't **need** to be published.

I like to keep the the workflows of commits to the main branch green. This allows me to quickly check if there are issues uploading a package or if there are other issues.

PR #6362, opened 2026-08-26
Add cgal-cpp

MMesch:
### Pre-submission Checks
- [ ] Package requires building for `emscripten-wasm32` platform (not a noarch package), in other words, the package requires compilation. .... To be honest not fully sure. The conda-forge recipe isn't noarch, but in theory this is a header only library that *maybe* could be noarch. I'm not sure whether there is an established way to do this for CPP libraries though, and also not how safe it is when CMAKE etc still run. Waiting on @IsabelParedes @DerThorsten 's opinion here.

### Recipe Structure
Added `recipes/recipes_emscripten/[package-name]/recipe.yaml` with proper structure:
- [x] `context` section with `version` (and optionally `name`)
- [x] `source` section with:
  - Source URL is valid and points to archive file (`.tar.gz`, `.tar.bz2`, `.tar.xz`, `.tgz`, or `.zip`)
  - Source URL contains `${{ version }}` template for version updates
  - SHA256 hash is correct (verified with `curl -sL <url> | sha256sum`)
  - Patches (if any) are included in `[package-name]/patches/` directory
- [x] `build` section with appropriate script/method
  - C++ packages: Uses `emcmake`/`emmake` or `emconfigure`/`emmake`
  - Build number is 0
  - If the script is longer than 3 lines, a *build.sh* is included
- [x] `requirements` section (build, host, run as needed)
- [x] `tests` section
  - C++ packages: Test executable or package_contents test

## PR Formatting
- [x] PR title follows format: `Add [package-name]``
  - Version being added/updated

## Package Details
- **Package Name**:  cgal-cpp
- **Version**: 6.0.3

## Build Notes

- header-only distribution of CGAL 6.0.3
- packaged under the same name as on conda-forge
- adding 6.0.3 because it's needed by a downstream package. Current newest is 6.2.
- since it's header only, I considered noarch packaging on conda-forge but unsure about it
- test compiles and runs under Node. It exercises 3 paths, one of them guarding against CGAL#9062 needing -fwasm-exceptions, which the emscripten-forge toolchain enables by default.

### DIFF
--- a/recipes/recipes_emscripten/cgal-cpp/recipe.yaml
+++ b/recipes/recipes_emscripten/cgal-cpp/recipe.yaml
@@ -12,9 +12,16 @@ source:
 
 build:
   number: 0
+  # CGAL 5+ is header-only, so the built artifact is arch-independent.
+  # Upstream conda-forge tracking: https://github.com/conda-forge/cgal-cpp-feedstock/issues/55
+  noarch: generic
 
 requirements:
   build:
+  # Even though nothing gets compiled, CGAL's CMakeLists calls
+  # project(CGAL LANGUAGES CXX), which triggers compiler detection at
+  # configure time. Pin one so it doesn't fall back to /usr/bin/c++.
+  - ${{ compiler('cxx') }}
   - cmake
   - ninja
   host:
--- a/recipes/recipes_emscripten/cgal-cpp/recipe.yaml
+++ b/recipes/recipes_emscripten/cgal-cpp/recipe.yaml
@@ -1,0 +1,58 @@
+context:
+  name: cgal-cpp
+  version: 6.0.3
+
+package:
+  name: ${{ name }}
+  version: ${{ version }}
+
+source:
+  url: https://github.com/CGAL/cgal/releases/download/v${{ version }}/CGAL-${{ version }}-library.tar.xz
+  sha256: 5c82166e6934397a523f2cb0aa35921d17e97f56fbc24e649a8627ef60a76a5c
+
+build:
+  number: 0
+
+requirements:
+  build:
+  - cmake
+  - ninja
+  host:
+  - boost-cpp
+  - gmp
+  - mpfr
+  run:
+  - boost-cpp
+  - gmp
+  - mpfr
+
+tests:
+- package_contents:
+    files:
+    - include/CGAL/version.h
+    - lib/cmake/CGAL/CGALConfig.cmake
+- script:
+  - build_tests.sh
+  requirements:
+    build:
+    - ${{ compiler('cxx') }}
+    - cmake
+    - ninja
+    - nodejs
+  files:
+    recipe:
+    - build_tests.sh
+    - tests/
+
+about:
+  homepage: https://www.cgal.org/
+  license: LGPL-3.0-or-later AND GPL-3.0-or-later
+  license_file:
+  - LICENSE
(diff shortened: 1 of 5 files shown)


MMesch:
Also added an upstream request: https://github.com/conda-forge/cgal-cpp-feedstock/issues/55

recipes/recipes_emscripten/cgal-cpp/recipe.yaml
```diff
@@ -0,0 +1,58 @@
...
+package:
+  name: ${{ name }}
+  version: ${{ version }}
+
+source:
+  url: https://github.com/CGAL/cgal/releases/download/v${{ version }}/CGAL-${{ version }}-library.tar.xz
+  sha256: 5c82166e6934397a523f2cb0aa35921d17e97f56fbc24e649a8627ef60a76a5c
+
+build:
+  number: 0
```
IsabelParedes, line 14: Why not make this noarch?

recipes/recipes_emscripten/cgal-cpp/recipe.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+requirements:
+  build:
+  - cmake
```
IsabelParedes, line 25: Even if there is no compilation, it would be safer to have a compiler listed

[This line](https://github.com/emscripten-forge/recipes/actions/runs/32991827166/job/98250963484?pr=6362#step:8:151) seems a bit concerning
```
-- Using /usr/bin/c++ compiler.
```

IsabelParedes (changes requested):
A couple of questions. Thanks!

IsabelParedes:
> Also added an upstream request: [conda-forge/cgal-cpp-feedstock#55](https://github.com/conda-forge/cgal-cpp-feedstock/issues/55)

Could you also add this link to the recipe directly? That way we could keep track of it more easily

recipes/recipes_emscripten/cgal-cpp/recipe.yaml
MMesch: did this now. I think it's genuinly noarch but it's dangerous for such a library because you basically need to analyze the cmake file to be sure?

recipes/recipes_emscripten/cgal-cpp/recipe.yaml
MMesch: did this

MMesch:
@IsabelParedes I think all your comments are addressed. Thanks!

recipes/recipes_emscripten/cgal-cpp/recipe.yaml
IsabelParedes: I thought you had already compared all the arch versions of this package from conda-forge, and they were all identical.

If the cmake produces identical headers for all architectures, then what is the danger?

IsabelParedes (approved):
Thank you!

recipes/recipes_emscripten/cgal-cpp/recipe.yaml
MMesch: I am not experienced enough with cmake. From my uninformed opinion and experience with other similar systems, I don't know if cmake could still bake in platform information realistically. Maybe not now but in an update? I'm potentially overly worried but the lack of testing for other platforms would make me nervous.

PR #6365, opened 2026-08-26
Add libboost-python 1.87.0

MMesch:
### Pre-submission Checks
- [x] Package requires building for `emscripten-wasm32` platform.

### Recipe Structure
- [x] `context` section with `version`
- [x] `source` section with valid URL, `${{ version }}` template, verified SHA256
- [x] `build` section with `build.sh` (invokes Boost b2)
- [x] `requirements` section (build/host/run)
- [x] `tests` section (`package_contents` + `emar t` archive inspection)

## PR Formatting
- [x] PR title follows format: `Add libboost-python`
- [x] Version being added: 1.87.0

## Package Details
- **Package Name**: libboost-python
- **Version**: 1.87.0

## Build Notes

Builds Boost.Python via the same b2 setup as `boost-cpp`, scoped to the python component via `bootstrap.sh --with-libraries=python`. Ships `libboost_python313-313.a` (matching cross-python 3.13). Boost headers come from `boost-cpp`.

Two things reviewers may want to flag:

1. `b2 install` is invoked with `|| true` and followed by a `test -f` on `libboost_python*.a`. Reason: b2's `install` still schedules unrelated Boost libs (container, graph, cobalt) that fail to cross-compile for emscripten, and `--with-X` and `--without-X` can't be mixed. Suggestions welcome for a cleaner scope.

2. Test uses `emar t` archive inspection rather than a compile-and-link test. A compile test would need `Python.h` and thus the cross-python venv, which the test env doesn't activate the way the build env does (`crossenv` missing at `$BUILD_PREFIX/bin/python`). Functional testing is planned via pytester once there's a downstream consumer.

Motivating downstream: WASM build of pygplates.


MMesch:
@IsabelParedes ready for review.

recipes/recipes_emscripten/libboost-python/recipe.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - ${{ compiler('cxx') }}
+    - cross-python_${{ target_platform }}
+    - python
+    - pip
+    host:
+    - ${{ pin_subpackage('libboost-python', exact=True) }}
+    - python
+    - pip
+    run:
+    - ${{ pin_subpackage('libboost-python', exact=True) }}
```
DerThorsten: on runtime we never depend  static libraries

recipes/recipes_emscripten/libboost-python/build_tests.sh
DerThorsten, line 1: I think we can drop all of this testing, since the testing via the actual example package is more robost

recipes/recipes_emscripten/libboost-python/example/pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "mymodule"
+version = "0.1.0"
```
DerThorsten: this version is not matching the pkgs version

recipes/recipes_emscripten/libboost-python/example/mymodule.cpp
DerThorsten, line 1: if you are brave, try throwing an exception from c++ and catching it on the python side

DerThorsten (changes requested):

recipes/recipes_emscripten/libboost-python/example/pyproject.toml
MMesch: changed

recipes/recipes_emscripten/libboost-python/example/mymodule.cpp
MMesch: seems to work!

recipes/recipes_emscripten/libboost-python/recipe.yaml
MMesch: removed. Thanks.

DerThorsten (approved):

PR #6376, opened 2026-08-27
Add qt6 multi-output

MMesch:
I'm trying to package Qt6 for `emscripten-wasm32` as a single multi-output recipe (qt6-main + qt6-svg + qt6-5compat) so the three modules stay version-locked, à la conda-forge. It's blocked, and I'm not fully sure why. Writing up what I see in case someone recognises it.

The build stops here:

```
Resolving build environment
  Platform: linux-64
  Specs:
   - qt6-main ==6.11.2 py313pl5321h3176c42_0
   ...

Error: × Failed to resolve dependencies
  ╰─▶ Cannot solve the request because of: No candidates were found
      for qt6-main ==6.11.2 py313pl5321h3176c42_0.
```

`py313pl5321h3176c42_0` is the hash of the **wasm** qt6-main I just built as an intra-recipe output. It obviously doesn't exist on `linux-64`. But this is the *build* env, which should solve against `linux-64/` and pick up conda-forge's native qt6-main (for `moc`, `rcc`, `uic`, `qt-configure-module`).

My working hypothesis: rattler-build auto-pins any dep reference to an intra-recipe output name to that output's exact `version + build-hash`, and it doesn't care that the section's platform is different from where that hash actually lives.

In a single-output recipe this never comes up. `qt6-main` in `build:` gets solved against `linux-64/` (conda-forge wins), and in `host:` / `run:` against `emscripten-wasm32/` (emscripten-forge wins). Two roles, one name, no conflict. Multi-output somewhere loses that platform-awareness for any name that matches one of its own outputs.

Things I've tried to escape the auto-pin, in order:

1. Swap `qt6-main` for `${{ pin_subpackage('qt6-main', exact=True) }}` in `host:` / `run:` while leaving `build:` as the plain name. Same error.
2. Channel-prefix the build dep as `conda-forge::qt6-main`. Same error, so the auto-pin ignores channel prefixes too.
3. Rename the real qtbase output to `qt6-main-libs` and add a trailing `qt6-main` metapackage (last in the outputs list) that just run-deps on it, hoping outputs are processed sequentially and earlier `build: qt6-main` refs wouldn't see the metapackage. Instead I got:
   ```
   Cycle detected in recipe outputs: qt6-5compat, qt6-main, qt6-main-libs, qt6-svg
   ```
   which suggests rattler resolves all output names in one pass before variant analysis, so forward-declaration doesn't help.

If the hypothesis is right, what would unblock this is either (a) making the auto-pin platform-aware, only firing for sections whose subdir matches the intra-recipe output's, or (b) some per-line escape hatch that says "this reference always goes to channels". Very open to being told I'm misreading it.


MMesch:
Converting to multi-output recipe that includes other qt packages.

MMesch:
@wolfv , is this potentially a rattler bug? Anything I missed or workaround?

recipes/recipes_emscripten/qt6/recipe.yaml
```diff
@@ -0,0 +1,132 @@
...
+  build:
+    script: build_qtbase.sh
+  requirements:
+    build:
+    - ${{ compiler('cxx') }}
+    - cmake
+    - ninja
+    - perl
+    - python
+    - qt6-main
```
IsabelParedes:
```suggestion
    - qt6-main ==${{ version }}
```

recipes/recipes_emscripten/qt6/recipe.yaml
```diff
@@ -0,0 +1,132 @@
...
+  build:
+    script: build_qtsvg.sh
+  requirements:
+    build:
+    - ${{ compiler('cxx') }}
+    - cmake
+    - ninja
+    - perl
+    - python
+    - qt6-main
```
IsabelParedes:
```suggestion
    - qt6-main ==${{ version }}
```

recipes/recipes_emscripten/qt6/recipe.yaml
```diff
@@ -0,0 +1,132 @@
...
+  build:
+    script: build_qt5compat.sh
+  requirements:
+    build:
+    - ${{ compiler('cxx') }}
+    - cmake
+    - ninja
+    - perl
+    - python
+    - qt6-main
```
IsabelParedes:
```suggestion
    - qt6-main ==${{ version }}
```

IsabelParedes:
> @wolfv , is this potentially a rattler bug? Anything I missed or workaround?

Not a rattler bug, rattler is working as expected.

recipes/recipes_emscripten/qt6/recipe.yaml
MMesch: done

recipes/recipes_emscripten/qt6/recipe.yaml
MMesch: done

MMesch:
Awesome thanks. I used your suggestion and added a few more packages. Probably not complete but enough for what I want to use it for: to build another package that relies on the base types.

recipes/recipes_emscripten/qt6/recipe.yaml
MMesch: done

MMesch:
@IsabelParedes I think this one is good now as well.

recipes/recipes_emscripten/qt6/build_qtbase.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ ! -f "${EMSDK}/.emscripten" ]; then
+    cat > "${EMSDK}/.emscripten" <<EOF
+LLVM_ROOT = '${EMSDK}/upstream/bin'
+BINARYEN_ROOT = '${EMSDK}/upstream'
+EMSCRIPTEN_ROOT = 'upstream/emscripten'
+NODE_JS = '${EMSDK_NODE}'
+COMPILER_ENGINE = NODE_JS
+JS_ENGINES = [NODE_JS]
+EOF
+fi
```
IsabelParedes, line 24: I wonder if this should be part of the activation script for the emscripten package, that way all packages use the same compiler configuration.

This is fine for now. But maybe we should explore this on the new `emscripten-6x` branch.

recipes/recipes_emscripten/qt6/recipe.yaml
```diff
@@ -0,0 +1,254 @@
...
+  target_directory: qt5compat
+- url: https://download.qt.io/official_releases/qt/6.11/${{ version }}/submodules/qtimageformats-everywhere-src-${{ version }}.tar.xz
+  sha256: cecd8900f34b6550076309bc94f62f828008b633a4239e0a08c86788f41001f8
+  target_directory: qtimageformats
+- url: https://download.qt.io/official_releases/qt/6.11/${{ version }}/submodules/qtwebchannel-everywhere-src-${{ version }}.tar.xz
+  sha256: feb3149758bda887f0c292656194812dd2c227595badc8fbdeb2fb2311c5575f
+  target_directory: qtwebchannel
+- url: https://download.qt.io/official_releases/qt/6.11/${{ version }}/submodules/qtwebsockets-everywhere-src-${{ version }}.tar.xz
+  sha256: 2bfe25c383c37f7690e3c59b02c9907d2744cf8527b1f3f1fc2c785b2ff3c9a3
+  target_directory: qtwebsockets
```
IsabelParedes: Currently, all of the sources are downloaded for every single recipe output. It would be more efficient to move each source to its corresponding output, that way each output only needs to download one source. And it would also eliminate the need for `target_directories`.

recipes/recipes_emscripten/qt6/recipe.yaml
```diff
@@ -0,0 +1,254 @@
...
+  - script:
+    - build_tests_qtwebsockets.sh
+    requirements:
+      build:
+      - ${{ compiler('cxx') }}
+    files:
+      recipe:
+      - build_tests_qtwebsockets.sh
+  about:
+    summary: Qt6 WebSockets (RFC 6455) client/server for Emscripten (WASM)
```
IsabelParedes:
```suggestion
    summary: Qt6 WebSockets (RFC 6455) client for Emscripten (WASM)
```

It might be misleading to claim the server works ([source](https://doc.qt.io/qt-6/wasm.html)).

> Note: [QWebSocketServer](https://doc.qt.io/qt-6/qwebsocketserver.html) is not supported due to browser limitations. Browsers restrict server-side socket functionality to ensure security within the web sandbox. Consequently, any functionality relying on [QWebSocketServer](https://doc.qt.io/qt-6/qwebsocketserver.html) for accepting incoming network connections cannot be utilized within the web environment.

IsabelParedes (changes requested):
Minor change, everything else looks good.

Thank you!

recipes/recipes_emscripten/qt6/build_qtbase.sh
MMesch: Makes sense to me but leaving this question to you ;)

recipes/recipes_emscripten/qt6/recipe.yaml
MMesch: done

recipes/recipes_emscripten/qt6/build_qtbase.sh
MMesch: Or do you want me to open an issue with this?

MMesch:
Addressed @IsabelParedes

recipes/recipes_emscripten/qt6/build_qtbase.sh
IsabelParedes: Could you please open the issue? that would be great

IsabelParedes (approved):
Thank you!

PR #6609, opened 2026-09-08
boost-cpp: split into conda package per Boost module

Alex-PLACET:
## Package Details
- **Package Name**: boost
- **Version**: 1.95

Use latest version of boost and create one package per components, even header-only. boost-headers groups all header-only packages.

Package build takes almost three hours, mostly because of the setup of the test environements for each packages (download of emscripten for each test env, sounds really inneficient)


IsabelParedes:
I'll wait until it's all green to review :)

Alex-PLACET:
@IsabelParedes it's green now :+1:

recipes/recipes_emscripten/boost-cpp/recipe.yaml
```diff
@@ -1,38 +1,6405 @@
...
+      script: build.sh
+    requirements:
+      build:
+        - ${{ compiler('cxx') }}
+      host:
+        - bzip2
+        - zlib
+    source:
+      - git: https://github.com/boostorg/boost.git
+        tag: boost-${{ version }}
```
IsabelParedes, line 22: The bot cannot update this. Please add a source with a url that has `${{ version }}`. The previous source was better.

The linter should have caught this. I'll look into that.

IsabelParedes (changes requested):
@Alex-PLACET , you are correct. This is tooooo many test files. Even GH has trouble rendering the page with changes.

I would suggest moving all the test files to an external repo. They do not need to be part of this recipe directory.

Then in the staging output, add a secondary source that points to that repo and puts all the tests in a `tests/` directory.

> Package build takes almost three hours, mostly because of the setup of the test environements for each packages (download of emscripten for each test env, sounds really inneficient)

You're correct. This is very inefficient. Since you already have a staging output, and with the second source you would have all of the tests there, you can compile all of the tests at that point. And you could even run all of the tests in staging (I haven't checked if `staging` supports a tests section, but probably).

Then for all of the other packages, checking the file contents will suffice.

Additionally, I don't see any `run_exports` or other type of version constraints. Is any `boost-*` package compatible with any other version of other `boost-*` packages?

Thanks!

recipes/recipes_emscripten/boost-cpp/recipe.yaml
Alex-PLACET: We can't use the archive which was used before because it don't have boostdeps program which is used to get the dependencies between the boost components

Alex-PLACET:
@IsabelParedes the point of having test for each package is also to verify that the package's depencies are correct too. Compiling all the tests with the staging do not have any benefit in my point of view.

Alex-PLACET:
run_exports added :+1:

IsabelParedes (approved):
Thank you!

Issue #6017, opened 2026-07-24
Packages ecl, fricas0-ecl overlap


mkoeppe:
```
Linking fricas0-ecl-26.5.5-h4599760_1
...
Linking ecl-26.5.5-h4599760_0
warning  libmamba [ecl-26.5.5-h4599760_0] The following files were already present in the environment:
    - COPYING
    - LICENSE
    - TAGS
    - asdf.fasc
    - bin/ecl.html
    - bin/ecl.js
    - bin/ecl.wasm
    - build-stamp
    - dpp.html
    - ecl/bytecodes.h
    - ecl/cache.h
    - ecl/config-internal.h
    - ecl/config.h
    - ecl/configpre-int.h
    - ecl/configpre.h
    - ecl/cons.h
```
https://github.com/passagemath/passagemath-jupyterlite-demo/actions/runs/29866894404/job/88757280173#step:5:231

wangyenshu:
You can remove ecl. fricas0-ecl can be used as ecl, but with more memory and some optional libraries.

PR #6611, opened 2026-09-08
Update openblas-experimental to develop @ 803f36bc

jjerphan:

---

## PR Formatting
  - Version being added/updated

## Package Details
- **Package Name**: `openblas-experimental`
- **Version**: `0.3.35.dev0` (build 1), snapshot of [OpenMathLib/OpenBLAS](https://github.com/OpenMathLib/OpenBLAS) `develop` @ [`803f36bc`](https://github.com/OpenMathLib/OpenBLAS/commit/803f36bcaabba43ddee84bf455dfcc24dc7aa0eb)

## Build Notes

This refreshes the `develop` snapshot from `632ef874` to `803f36bc` (64 commits newer) and rebases the emforge patch series onto it. The rebased branch is [`v0.3.35.dev0-emforge`](https://github.com/jjerphan/OpenBLAS/tree/v0.3.35.dev0-emforge) at https://github.com/jjerphan/OpenBLAS.

The package version string is unchanged, so the build number is bumped to 1 rather than reset.

### Patches now upstream (dropped)

Three of the fourteen patches have landed upstream and were removed:

- `0006-Int-to-void-return-types-of-lapack-functions-used-by` — superseded by OpenBLAS [#6017](https://github.com/OpenMathLib/OpenBLAS/pull/6017)
- `0013-wasm-guard-KERNEL-assignments-with-ifndef` — OpenBLAS [#5996](https://github.com/OpenMathLib/OpenBLAS/pull/5996)
- `0014-wasm-use-relaxed-simd-madd-for-v_muladd` — OpenBLAS [#6001](https://github.com/OpenMathLib/OpenBLAS/pull/6001)

### Patches reduced

- `0002-Skip-linktest` no longer touches the WASM code path of `exports/Makefile`, since upstream added an `ifneq ($(ARCH),wasm)` guard there ([#6007](https://github.com/OpenMathLib/OpenBLAS/pull/6007)). It now only disables linktest for the remaining compiler branches and the `test` target.
- `0005-Return-void-not-int` shrinks from ~1170 changed lines to ~40: upstream converted the reimplemented LAPACK routines to `void`, so only `xerbla` and the OpenBLAS extension declarations that upstream left returning `int` remain.
- `0004-Add-dummy-length-arguments-for-char-arguments` was rebased onto the new signatures, so the Fortran hidden-length arguments are now added to `void OPENBLAS_API(...)` declarations instead of the old `int BLASFUNC(...)` ones.

The remaining 11 patches were renumbered, and the patch-range comments in `recipe.yaml` and `build.sh` were updated accordingly.

### Local verification

Built locally with `pixi run -e rattler-build-env build-emscripten-wasm32-pkg recipes/recipes_emscripten/openblas-experimental`:

- All 11 patches apply without fuzz.
- `utest` and the CBLAS `ctest` levels 1-3 (real and complex, column- and row-major) pass under Node.
- The packaged `test_openblas.py` passes in Chromium, Firefox and WebKit.


MMesch (approved):
LGTM. Waiting for build to run though.

PR #6664, opened 2026-09-09
qt6: enable wasm-exceptions + JSPI versions

MMesch:
Three commits:

1. **\`qt6: enable wasm-exceptions\`** — adds \`-feature-wasm-exceptions\` to qtbase configure, matching the emscripten-forge convention of always using \`-fwasm-exceptions\` on wasm.
2. **\`qt6: add JSPI variant outputs (qt6-main-jspi, qt6-5compat-jspi)\`** — new sibling packages rebuilt with \`QT_FEATURE_wasm_jspi=ON\` so Qt's event dispatcher can yield to the browser via JavaScript Promise Integration. Distinct package names; \`run_constraints\` prevent jspi/vanilla siblings from co-installing (they share file paths).
3. **\`qt6: extend JSPI variants to svg, imageformats, webchannel, websockets\`** — same pattern for the remaining Qt6 submodules so downstream apps needing any Qt6 submodule under JSPI don't hit file-conflicts.

## Build notes

- Why JSPI needs to be opt-in
  - Kernels on this channel (xeus-python, pyjs, etc.) don't yet support JSPI. Making JSPI the qt6 default would break Qt-based Python extensions loaded into those kernels.
  - Browser floor is also higher: Chrome 137+, Firefox 141+, Safari 18.4+.
- Opt-in via distinct package names keeps existing consumers untouched. Existing \`qt6-main\` consumers see nothing change except the wasm-exceptions ABI improvement.

### DIFF
--- a/recipes/recipes_emscripten/qt6/recipe.yaml
+++ b/recipes/recipes_emscripten/qt6/recipe.yaml
@@ -21,8 +21,6 @@ outputs:
     - ${{ compiler('cxx') }}
     - cmake
     - ninja
-    - perl
-    - python
     - qt6-main == ${{ version }}
     host:
     - zlib
@@ -66,8 +64,6 @@ outputs:
     - ${{ compiler('cxx') }}
     - cmake
     - ninja
-    - perl
-    - python
     - qt6-main == ${{ version }}
     host:
     - ${{ pin_subpackage('qt6-main', exact=True) }}
@@ -105,8 +101,6 @@ outputs:
     - ${{ compiler('cxx') }}
     - cmake
     - ninja
-    - perl
-    - python
     - qt6-main == ${{ version }}
     host:
     - ${{ pin_subpackage('qt6-main', exact=True) }}
@@ -143,8 +137,6 @@ outputs:
     - ${{ compiler('cxx') }}
     - cmake
     - ninja
-    - perl
-    - python
     - qt6-main == ${{ version }}
     host:
     - ${{ pin_subpackage('qt6-main', exact=True) }}
@@ -180,8 +172,6 @@ outputs:
     - ${{ compiler('cxx') }}
     - cmake
     - ninja
-    - perl
-    - python
     - qt6-main == ${{ version }}
     host:
     - ${{ pin_subpackage('qt6-main', exact=True) }}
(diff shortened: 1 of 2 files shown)


recipes/recipes_emscripten/qt6/recipe.yaml
```diff
@@ -240,6 +240,260 @@ outputs:
...
+# from their event loop to the browser (all of them, effectively) get proper
+# JSPI-based suspension. Distinct package names so the solver never picks
+# them unless a downstream explicitly asks; run_constraints prevent a jspi
+# variant from ever being installed alongside its vanilla counterpart (they
+# install identical file paths).
+- package:
+    name: qt6-main-jspi
+    version: ${{ version }}
+  source:
+    url: https://download.qt.io/official_releases/qt/6.11/${{ version }}/submodules/qtbase-everywhere-src-${{ version }}.tar.xz
```
IsabelParedes: Maybe it would be good to create a context variable for `6.11`. The bot will not update this when bumping versions

MMesch: yes!

recipes/recipes_emscripten/qt6/recipe.yaml
```diff
@@ -1,15 +1,18 @@
 context:
   version: 6.11.2
+  # Major.minor derived from version so the version-bump bot updates the
+  # URL prefix (qt/6.11/) automatically when version rolls to 6.12.x etc.
+  major_minor: ${{ (version | split('.'))[0] }}.${{ (version | split('.'))[1] }}
```
IsabelParedes, line 5: Nice!

IsabelParedes (changes requested):
Please remove `perl` and `python` from `build.requirements`.

Currently all of the packages get the substrings <mark>**py313pl5321**</mark>h3176c42_1, (from python and perl).

Perl is not a build requirement
- https://www.qt.io/blog/perl-is-out-of-the-list-of-dependencies

And `python` is a run dependency for the c/cxx compiler, so it automatically gets installed. Removing the redundant python would remove the `py313` from the packages build strings, to not give the impression that they are python packages.

Thanks!

MMesch:
done @IsabelParedes ! Good catch, thanks.

IsabelParedes (approved):
Thank you!

PR #6738, opened 2026-09-14
Add sqlitebrowser-experimental

MMesch:
## Package Details

- **Package Name**: sqlitebrowser
- **Version**: 3.13.99 (master branch @ \`4a7359d5\`, 2026-09-08)

Packages [DB Browser for SQLite](https://sqlitebrowser.org/) — the widely-used GUI for SQLite databases — cross-compiled to wasm. Routed to \`emscripten-forge-4x-experimental\`.

First 50K-line real-world Qt6 widget app on the channel. Depends on \`qt6-main-jspi\` + \`qt6-5compat-jspi\` (added in #6664) plus \`sqlite\`. Installs the four wasm artifacts (\`.wasm\`, \`.js\`, \`qtloader.js\`, and Qt's auto-generated \`.html\`) under \`share/sqlitebrowser/\`. Load via the [qt-wasm-runner](https://github.com/MMesch/qt-wasm-runner) or any static host.

## Why master, not v3.13.1

v3.13.1 (Oct 2024) is Qt5-only. Qt6 support landed on master afterwards via the \`QT_MAJOR\` cmake variable. We pin to a specific master commit (\`4a7359d5\`, 2026-09-08) that has it. Once upstream cuts v3.14.x with Qt6 support, we can retag to a stable release.

### Pre-submission
- [x] Requires building for \`emscripten-wasm32\` (cross-compiled C++)

### Recipe Structure
- [x] \`context\` with \`version\` + \`commit\` (pinning master)
- [x] \`package\` section with name/version templated
- [x] \`source\`: github archive URL of the pinned commit; SHA256 verified; 7 patches under \`patches/\` (see below)
- [x] Build script in \`build.sh\` (~40 lines: sets \`QT_HOST_PATH\`, synths \`.emscripten\`, invokes cmake with \`QT_MAJOR=Qt6\` + \`sqlcipher=OFF\` + explicit SQLite paths, ninja, then copies four wasm artifacts to \`share/sqlitebrowser/\`); build number 0
- [x] Requirements: \`cxx\`, \`cmake\`, \`ninja\`, \`qt6-main\` (build); \`qt6-main-jspi\`, \`qt6-5compat-jspi\`, \`sqlite\` (host)
- [x] Tests: \`package_contents\` verifying wasm artifacts
- [x] \`about\` section with license (\`MPL-2.0 OR GPL-3.0-or-later\`), homepage, summary

## Patches

Seven wasm-specific patches carry DB Browser through the build + runtime:

1. **\`0001-include-qsslerror-for-wasm.patch\`** — \`RemoteNetwork.h\` uses \`QList<QSslError>\` but only includes \`QSslConfiguration\`; on wasm Qt the transitive include chain doesn't reach \`QSslError\`. One-line explicit include.
2. **\`0002-stub-remote-network-on-wasm.patch\`** — Qt-wasm's \`QtNetwork\` ships without SSL and without \`QNetworkProxyFactory::systemProxyForQuery\`. Stub the (non-critical) \`RemoteNetwork\` class — it powers cloud sync + update-check, neither meaningful in a browser tab.
3. **\`0003-use-qt-add-executable-for-wasm.patch\`** — DB Browser calls plain \`add_executable(sqlitebrowser)\` at line 16, before \`find_package(Qt6)\` at line 63, so \`qt_add_executable\` can't be substituted. Instead, load Qt's \`QtPublicWasmToolchainHelpers.cmake\` early and call \`qt6_finalize_target()\` at the end — applies the same wasm-specific link flags (\`embind\`, \`MODULARIZE=1\`, \`EXPORT_NAME\`, memory sizes) that \`qt_add_executable\` would have.
4. **\`0004-scintilla-instantiate-runstyles-long-for-wasm.patch\`** — bundled Scintilla's explicit template instantiations of \`RunStyles<ptrdiff_t, X>\` are gated by \`#if PTRDIFF_MAX != INT_MAX\`, but on wasm32 \`ptrdiff_t\` is \`long\` (a distinct type from \`int\` despite same size). Add unconditional \`<long, int>\` and \`<long, char>\` instantiations for wasm.
5. **\`0005-wasm-file-open-via-browser-content-api.patch\`** — the traditional \`QFileDialog::getOpenFileName\` on wasm can't bridge browser files into MEMFS. Route the empty-filename branch through \`QFileDialog::getOpenFileContent\` (wasm-native, delivers bytes via callback), write to a MEMFS scratch path, then recursively open. Everything downstream in sqlite3 works unchanged.
6. **\`0006-wasm-file-save-via-browser-download.patch\`** — mirror of 0005 in the save direction: serialise the current DB via \`db.saveAs()\` into a MEMFS scratch, read the bytes, hand to \`QFileDialog::saveFileContent()\` which triggers a browser download.
7. **\`0007-wasm-reroute-file-new-to-in-memory.patch\`** — "New Database" via browser save-file-picker returns a one-shot download handle, not a writable path, leaving DB Browser in a "no database opened" state. Route \`fileNew()\` on wasm to \`fileNewInMemoryDatabase()\` (a working action DB Browser already has) so both menu entries do the right thing.

All seven are \`#ifdef __EMSCRIPTEN__\`-guarded and no-ops on non-wasm builds; potentially upstream-mergeable as-is.

## Verified working

- Extract package with rattler-build → serve via [qt-wasm-runner](https://github.com/MMesch/qt-wasm-runner) → Chrome 148
- New In-Memory Database → create tables → add rows → run SQL queries → Save As (downloads valid .sqlite roundtrippable to native DB Browser)
- Open Database → pick local .sqlite via browser file picker → tables visible, queries work

## Build Notes

- \`sqlcipher=OFF\` — not shipping encrypted DB support.
- No dynamic side-module deps; sqlite is statically linked. Package size ~15 MB.


MMesch:
merging to experimental without external review.

PR #6752, opened 2026-09-15
fix gap

wangyenshu:

- [x] Or reset build number to 0 if updating the package to a newer version

---

## PR Formatting
  - Version being added/updated

## Package Details
- **Package Name**: gap
- **Version**: 4.16.1


wangyenshu:
Screenshot29

I can confirm that gap works after this fix.

wangyenshu:
passagemath-gap test says
```
=================================== FAILURES ===================================
_________________________ test_import_passagemath_gap __________________________
    def test_import_passagemath_gap():
>       import passagemath_gap
E       ModuleNotFoundError: No module named 'passagemath_gap'
test_passagemath_gap.py:5: ModuleNotFoundError
_________________________________ test_libgap __________________________________
    def test_libgap():
>       import passagemath_gap
E       ModuleNotFoundError: No module named 'passagemath_gap'
test_passagemath_gap.py:9: ModuleNotFoundError
=========================== short test summary info ============================
```
I guess that this is not the gap's issue.

IsabelParedes:
> I guess that this is not the gap's issue.

Something seems to be broken with python tests. Investigating...

recipes/recipes_emscripten/gap/build.sh
```diff
@@ -35,6 +38,37 @@ fi
     make -j8
 )
 
+AUX_BUILD=$PWD/extern/emscripten/build
```
mkoeppe: Can you explain for what purpose you need to re-introduce these aux builds of packages that are already in emscripten-forge?

wangyenshu: Because gap's emscripten build script uses that. I will try to keep the changes minimal.

IsabelParedes:
The issue with the python tests is fixed now, if you'd like to test `passagemath-gap` again :)

IsabelParedes:
```
error while evaluating main file: Error: Dynamic linking error: cannot resolve symbol invoke_viii
```

This generally indicates that `-fexceptions` was used somewhere, but everything should use `-fwasm-exceptions` instead.

wangyenshu:
> ```
> error while evaluating main file: Error: Dynamic linking error: cannot resolve symbol invoke_viii
> ```
> 
> This generally indicates that `-fexceptions` was used somewhere, but everything should use `-fwasm-exceptions` instead.

But fwasm is not compatible with asyncify, which is required by gap's garbage collecting mechanism.

recipes/recipes_emscripten/gap/build.sh
IsabelParedes: But is it necessary? What is the issue with using the `zlib` and `gmp` packages?

recipes/recipes_emscripten/passagemath-gap/test_passagemath_gap.py
IsabelParedes, line 1: Please restore the changes in this file.

IsabelParedes (changes requested):
Since the changes to `gap` break `passagemath-gap`, I would like to get @mkoeppe opinion before merging.

Thank you!

recipes/recipes_emscripten/gap/build.sh
wangyenshu: Oh. I am careless. I will fix that.

wangyenshu:
Is it possible for the passagemath-gap test to work without -fwasm?

recipes/recipes_emscripten/passagemath-gap/test_passagemath_gap.py
wangyenshu: Finished.

recipes/recipes_emscripten/gap/build.sh
wangyenshu: Fixed.

mkoeppe:
> Since the changes to `gap` break `passagemath-gap`, I would like to get @mkoeppe opinion before merging.

Unfortunately I don't understand the constraints of this platform well enough to comment... What's going on in the dynamic linking of the Python modules that is failing here, and why is dynamic linking otherwise generally avoided in the emscripten-forge recipes?

wangyenshu:
> > Since the changes to `gap` break `passagemath-gap`, I would like to get @mkoeppe opinion before merging.
> 
> Unfortunately I don't understand the constraints of this platform well enough to comment... What's going on in the dynamic linking of the Python modules that is failing here, and why is dynamic linking otherwise generally avoided in the emscripten-forge recipes?

invoke_viii should be relevant to the JavaScript-based exception handling (-fexceptions). Most of the recipe here is built with WebAssembly-based exception handling (-fwasm). That should be the reason for failing. However, gap relies on asyncify flag which is not compatible with -fwasm.

recipes/recipes_emscripten/passagemath-gap/test_passagemath_gap.py
IsabelParedes: Sorry, I didn't clarify enough.

What I meant is to completely restore the changes so that it matches `main` exactly. Before there were 2 extra empty lines. This might seem insignificant, but the CI works by checking for **any** changes in a recipe directory. So having 2 extra empty lines or a new test (like below) will trigger a rebuild of `passagemath-gap`, and later it will fail to upload because the build number has not been bumped.

But this PR is about `gap` and not `passagemath-gap`, so there shouldn't be any changes to any of the `passagemath-gap/` files. Unless that's the intention, but then please bump the build number.

Thanks!

wangyenshu: Thanks for clarifying. I have restored passagemath-gap.

IsabelParedes (approved):
Thank you!

If this causes issues for `passagemath-gap`, we can pin it to a previous build or split the `gap` package as needed.

PR #6786, opened 2026-09-16
Revert boost recipe split and use version 1.92

Alex-PLACET:
## Package Details
- **Package Name**: boost-cpp 
- **Version**: 1.92


recipes/recipes_emscripten/boost-cpp/recipe.yaml
```diff
@@ -1,6723 +1,38 @@
...
-      run_exports:
-        - ${{ pin_subpackage('boost-geometry', upper_bound='x.x') }}
-    tests:
-      - script:
-          - bash tests/build_tests_cmake.sh geometry
-        requirements:
-          build:
-            - ${{ compiler("cxx") }}
+build:
+  number: 0
```
Alex-PLACET: should I increase it ? @IsabelParedes

IsabelParedes: yes, please

recipes/recipes_emscripten/boost-cpp/recipe.yaml
```diff
@@ -1,6723 +1,38 @@
 context:
   version: 1.92.0
-  build_number: 0
+  filename: boost_${{ version | replace('.', '_') }}.tar.bz2
```
IsabelParedes: To pass the lint check, could you remove the `.tar.bz2` and simply put it at the end of the url?
Thanks!

IsabelParedes (changes requested):
It does not seem necessary to revert the recipe. It would be good to have a split recipe with all the boost modules.

Thank you!

PR #6790, opened 2026-09-16
Update all recipes which have have a dependency on boost-cpp

Alex-PLACET:


recipes/recipes_emscripten/python-symengine/recipe.yaml
```diff
@@ -10,7 +10,7 @@ source:
   sha256: 3e79d39af5e9f024cd4b8d1372314ac2b83cace7d49cdbad482b4aec5b04c37b
 
 build:
-  number: 2
+  number: 4
```
IsabelParedes, line 13:
```suggestion
  number: 3
```

recipes/recipes_emscripten/cgal-cpp/recipe.yaml
```diff
@@ -9,12 +9,11 @@ package:
...
   url: https://github.com/CGAL/cgal/releases/download/v${{ version }}/CGAL-${{ version }}-library.tar.xz
   sha256: 8e7ce1e30e8ffcb3e69d153c3a561bb1f38bbb4becd8e6544cbaa0fb2889af62
+  patches:
+  - patches/keep-boost-find-module-mode.patch
 
 build:
-  number: 0
-  # CGAL 5+ is header-only, so the built artifact is arch-independent.
-  # Upstream conda-forge tracking: https://github.com/conda-forge/cgal-cpp-feedstock/issues/55
-  noarch: generic
```
IsabelParedes, line 17: Any reason for removing this?

recipes/recipes_emscripten/arrow/recipe.yaml
```diff
@@ -327,6 +355,7 @@ outputs:
   - package:
       name: r-arrow
```
IsabelParedes, line 356: We now maintain 2 variants of all R packages, one for R v4.5 and another for R v4.6.

I think it might be better to move r-arrow to its own recipe, that way all of the arrow packages do not need to be built twice.

But that can be done later. This is fine for now.

IsabelParedes (changes requested):
Changes to the host environments do not seem necessary.

Thanks!

PR #6862, opened 2026-09-21
sqlitebrowser-experimental: run RowLoader synchronously on wasm

MMesch:
## Template B: updating a package

- [x] ⚠️ Bump build number if the version remains unchanged (0 → 1)

## Package Details
- **Package Name**: sqlitebrowser-experimental
- **Version**: 3.13.99 (unchanged)

## Build Notes

DB Browser's `SqliteTableModel` loads rows through a `RowLoader` that inherits `QThread` and uses `std::async` / `std::mutex` / `std::condition_variable` to fetch rows on a worker thread. Qt-wasm doesn't ship pthreads, so `QThread::run()` never fires and `std::async`'s async policy is a no-op: opening a table shows an empty view because rows are queued to a worker that never runs.

New patch `0008-wasm-row-loader-synchronous.patch` runs `triggerFetch()` and `triggerRowCountDetermination()` directly on the calling thread when `__EMSCRIPTEN__` is defined. Signals still fire via `Qt::QueuedConnection` so UI updates land on the next event-loop tick. Native builds unchanged.

Verified locally: opens 10k / 100k / 1M-row demo databases and browses tables correctly.


MMesch:
bypass merge to experimental.

PR #6864, opened 2026-09-21
Update qt6-main (build 3): install-time and focus fixes

MMesch:

- [x] ⚠️ Bump build number if the version remains unchanged (2 → 3)

## PR Formatting

  - Version being added/updated

## Package Details

- **Package Name**: qt6-main
- **Version**: 6.11.2 (unchanged, build 2 → 3)

## Build Notes

Four fixes, each in its own commit and self-contained.

### 1. `qt6: rewrite install-time absolute paths in wrapper scripts`

Qt cross-install writes shell wrappers (`qmake6`, `qtpaths6`, `qt-cmake`) into `$PREFIX/bin` that hardcode the build-time absolute path of the native tools in `$BUILD_PREFIX`. On any other machine those paths don't exist and the wrappers fail with `No such file or directory`. Post-install `sed` rewrites the baked prefix to `${QT_HOST_PATH:?...}`. Downstream cross-compile flows already set `QT_HOST_PATH`, so wrappers work with no additional setup.

### 2. `qt6: avoid SIGPIPE in qt6-main test script`

`build_tests_qtbase.sh` uses `set -euo pipefail`. `echo "${MEMBERS}" | head` and `... | grep '\.o$' | head -1` hit SIGPIPE (exit 141) when `MEMBERS` is large enough that `head` closes the pipe early. Replaced with here-strings and `grep -m1` so no pipe is involved.

### 3. `qt6: patch wasm plugin to release focus on tab blur`

`QWasmScreen` never listened for browser tab focus loss, so on tab return `QGuiApplication::focusWindow()` still pointed at the app's window and the `PointerDown` path in `pointerEventForWindow` short-circuited `requestActivate()`. Result: keyboard was dead until a menu popup forced a full activation cycle. Adds a `blur` listener on `window` that calls `handleFocusWindowChanged(nullptr)` when the tab loses focus.

Uses `window.blur` rather than `focusout` on the shadow container because focusout leaks during in-app focus transitions (menu popups, dialog opens) and clearing Qt's focus mid-activation causes other regressions.

### 4. `qt6: patch wasm plugin to activate top-level windows on show`

`QWasmWindow::setVisible(true)` toggled DOM visibility but never requested window activation, so a re-shown non-modal `QDialog` (e.g. `dialog->show()` after `dialog->hide()`) appeared with the `inactive` decoration class, dimmed, without keyboard focus. Adds `window()->requestActivate()` on the `visible` transition, restricted to `Qt::Dialog` type that accepts focus, gated on `focusWindow() != window()` so it's a no-op for already-active windows and doesn't touch tooltips/popups (which have their own grab-based handling) or `WA_ShowWithoutActivating` background windows.


DerThorsten (approved):
lgtm

PR #6882, opened 2026-09-22
Update sqlitebrowser-experimental: run Execute-SQL synchronously on wasm

MMesch:

- [x] ⚠️ Bump build number if the version remains unchanged (2 → 3)

## PR Formatting

  - Version being added/updated

## Package Details

- **Package Name**: sqlitebrowser-experimental
- **Version**: 3.13.99 (unchanged, build 2 → 3)

## Build Notes

Follow-up to #6862. After the `RowLoader` sync patch made the table browser work on wasm, the Execute-SQL tab was the other threading-dependent surface still broken. This PR fixes that.

DB Browser's Execute-SQL runs the query on a `QThread` subclass (`RunSql`) and blocks on `std::condition_variable::wait()` between statements, waiting for the main thread to call `startNextStatement()` from a signal handler. Qt-wasm without pthreads never runs `QThread::run()`, so hitting Execute or Ctrl+Enter does nothing.

New patch `0009-wasm-run-sql-synchronous.patch`:
- Overrides `QThread::start()` on wasm to run the statement loop synchronously on the calling thread.
- Skips the inter-statement `cv.wait()` — there's no other thread that could call `startNextStatement()`.
- Fires `finished` via `QMetaObject::invokeMethod(this, "finished", Qt::QueuedConnection)` at the end so the UI slot re-enables the buttons and clears the read-only editor. (Direct `emit finished()` fails to compile because `QThread::finished` is a `QPrivateSignal`.)

Signals still fire via `Qt::QueuedConnection`, so slot handlers update the UI on the next event-loop tick — matching the threaded contract as far as the app can tell.

Verified locally: Execute All, Execute Line, Execute Selection all work; row results appear in the results pane; buttons re-enable after execution.

## Threading survey

After this PR, all threading uses in DB Browser's source tree are patched or benign:
- `RowLoader` — patched by #6862.
- `RunSql` — patched by this PR.
- `sqlitedb.cpp` `QThread::currentThread()` check — evaluates as same-thread on wasm, no behavior change.
- `sqlitetablemodel.cpp` `#include <QtConcurrent/...>` — dead include, no actual `QtConcurrent::run()` call.


MMesch:
bypassing to experimental

